PR #13, opened 2020-04-05
Use modules, add prettier

danwang:
r? @rbtying 

Updated tsconfig to allow imports from node modules (React/ReactDOM) and local component files. I moved some code into separate files:
- Error, the simplest component I could find
- beep function
- Trump component, as well as changing ITrump into a disjoint union to prevent the impossible case
- FriendSelect

Added prettier, can be run with `yarn prettier --fix`

### DIFF
--- a/frontend/src/FriendSelect.tsx
+++ b/frontend/src/FriendSelect.tsx
@@ -1,0 +1,63 @@
+import * as React from 'react';
+import {ITrump} from './types';
+
+type FriendSelection = {
+  card: string;
+  skip: number;
+};
+type Props = {
+  friend: FriendSelection;
+  trump: ITrump;
+  num_decks: number;
+  onChange: (input: FriendSelection) => void;
+};
+
+const FriendSelect = (props: Props) => {
+  const handleChange = (
+    transform: (e: HTMLSelectElement) => Partial<FriendSelection>,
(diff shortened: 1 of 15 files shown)


frontend/src/FriendSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+type Props = {
+  friend: FriendSelection;
+  trump: ITrump;
+  num_decks: number;
+  onChange: (input: FriendSelection) => void;
+};
+
+const FriendSelect = (props: Props) => {
+  const handleChange = (
+    transform: (e: HTMLSelectElement) => Partial<FriendSelection>,
```
rbtying, line 17: o.O what is this magic !?

frontend/static/index.html
```diff
@@ -7,12 +7,6 @@
     </head>
     <body>
         <div id="root"></div>
-        <script crossorigin src="https://unpkg.com/react@16/umd/react.development.js"></script>
```
rbtying, line 10: hm. I think there is a way to tell Webpack that React is available as an external global, so that it can be preloaded from CDN? I don't know if the tree-shaking stuff makes the effect of loading it ourselves pointless...

(nonblocking)

danwang: I did this to get unminified errors in development. As far as I know, the previous way using `externals` is the right way to load from CDN

PR #23, opened 2020-04-05
Add reset button

### LINKED ISSUE #20: Button to reset an ongoing game to Initialize

rbtying:
r? @danwang 

Add a reset button for when the game should be set back to the initialize state. Fixes #20

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -1365,6 +1365,23 @@ function renderUI() {
             <div className={state.four_color ? 'four-color' : ''}>
               <Errors errors={state.errors} />
               <div className="game">
+                {state.game_state.Initialize ? null : (
+                  <a
+                    href={window.location.href}
+                    className="reset-link"
+                    onClick={(evt) => {
+                      evt.preventDefault();
+                      if (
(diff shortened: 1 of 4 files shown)


frontend/src/index.tsx
```diff
@@ -1424,6 +1424,21 @@ function renderUI() {
...
+            {state.game_state.Initialize ? null : (
+              <a
+                href={window.location.href}
+                className="reset-link"
+                onClick={(evt) => {
+                  evt.preventDefault();
+                  if (window.confirm('Do you really want to reset the game?')) {
+                    send({Action: 'ResetGame'});
+                  }
+                  renderUI();
```
danwang, line 1436: A next step after this might be to make the entire application computable from the input state. So, for example:

- Create a top-level component that maintains the state
- Any time we receive a websocket input, use `setState` on it
- Here, fire-and-forget the `send`, and remove the imperative render

rbtying: yeah, that makes sense.

Most of the components are structured that way, but the things which directly render the state global are not (mostly; I didn't know how to)

PR #26, opened 2020-04-06
Create React-based application state

danwang:
r? @rbtying 
This creates a place for state to be stored within React state. Recommended view order:

1. `src/Chat.tsx` was moved out of `index.tsx`
2. `src/State.ts` contains types for a "state definition". Right now it just includes two methods, `loadDefault` and `persist`. There's a builder `noPersistence` for state that doesn't intend to be written to localStorage
3. `src/localStorageState.ts` implements State using localStorage as a persistence layer
4. `src/state/Settings.ts` uses localStorageState to create a State object for the settings hash
5. `src/AppStateProvider.tsx` implements AppState, which is the same as the global State in the beginning. It uses React.Context to provide two values: `state` and `updateState` anywhere within the render tree. There are two exports:
  - `AppStateConsumer` is used to read the context values
  - `AppStateProvider` is used once in the root of the app to read and provide state
6. `SettingsButton` now uses AppState directly to retrieve the settings.

Within `index.tsx`, we're creating the new React-based AppStateProvider, but **only settings** are being read/written from/to this. Everything else continues to use the global state + `renderUI` combo. I removed the settings from top-level state and things still work.


frontend/src/Chat.tsx
```diff
@@ -0,0 +1,54 @@
...
+const Chat = (props: Props) => {
+  const [draft, setDraft] = React.useState<string>('');
+  const anchor = React.useRef(null);
+
+  // TODO: messages is mutable, so can't use reference equality to check for
+  // new state. After making messages immutable, update to only rely on
+  // props.messages
+  React.useEffect(() => {
+    anchor.current?.scrollIntoView({block: 'nearest', inline: 'start'});
+  }, [props.messages.length]);
```
rbtying, line 18: hm. Can we do a full array-equality check for `props.messages` here?

Not only is messages mutable, the onmessage handler bounds its length to 100 entries, so this would stop scrolling at some point?

Incidentally, how would we make messages immutable in this case?

danwang: in the future, we can do something like this: https://github.com/rbtying/shengji/compare/danwang/state2?expand=1#diff-26b53db46abef847e920b00088370589R19

It computes a new array of messages and passes the new thing to the state to be stored.

When we rerender, we can use pointer equality to see whether or not it's different, and components will get the new copy passed down

danwang:
r? @rbtying 

Updated to work work the mutable array + 100 limit here: 9042e603213f129703d602739ea76b4a84f9eadb

PR #28, opened 2020-04-06
Migrate state to application state

danwang:
Moves all websocket-initiated state to AppState

- changed the behavior of errors -- instead of removing individual error messages, all disappear at the same time after 5000ms of no new errors
- Added WebsocketProvider, which does two things:
  - Sets up connection to websocket, binds listeners to it, and updates AppState as needed
  - Creates a `send` and allows users to read from it. Nobody does this yet; it's just set as a global
- Removed all ReactDOM.render except 1, put it into a bootstrap() function

### DIFF
--- a/frontend/src/Errors.tsx
+++ b/frontend/src/Errors.tsx
@@ -1,4 +1,6 @@
 import * as React from 'react';
+import {AppStateConsumer} from './AppStateProvider';
+import Timeout from './Timeout';
 
 type Props = {
   errors: string[];
@@ -6,6 +8,11 @@ type Props = {
 
 const Errors = (props: Props) => (
   <div className="errors">
+    <AppStateConsumer>
+      {({updateState}) => (
+        <Timeout timeout={5000} callback={() => updateState({errors: []})} />
+      )}
+    </AppStateConsumer>
     {props.errors.map((err, idx) => (
       <p key={idx}>
         <code>{err}</code>
--- a/frontend/src/WebsocketProvider.tsx
+++ b/frontend/src/WebsocketProvider.tsx
@@ -1,0 +1,63 @@
+import * as React from 'react';
+import {AppState, AppStateConsumer} from './AppStateProvider';
+import websocketHandler from './websocketHandler';
+
+type Context = {
+  send: (value: any) => void;
+};
+
+const WebsocketContext = React.createContext<Context>({
+  send: () => {},
+});
+
+export const WebsocketConsumer = WebsocketContext.Consumer;
+
+type Props = {
+  state: AppState;
+  updateState: (state: Partial<AppState>) => void;
+  children?: React.ReactNode;
+};
+const WebsocketProvider = (props: Props) => {
+  const {state, updateState, children} = props;
+  const [websocket, setWebsocket] = React.useState<WebSocket | null>(null);
+
+  React.useEffect(() => {
+    const uri =
+      (location.protocol === 'https:' ? 'wss://' : 'ws://') +
+      location.host +
+      location.pathname +
+      (location.pathname.endsWith('/') ? 'api' : '/api');
+
+    const ws = new WebSocket(uri);
+    setWebsocket(ws);
+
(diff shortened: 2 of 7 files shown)


frontend/src/Errors.tsx
```diff
@@ -1,11 +1,18 @@
...
 
 type Props = {
   errors: string[];
 };
 
 const Errors = (props: Props) => (
   <div className="errors">
+    <AppStateConsumer>
+      {({updateState}) => (
+        <Timeout timeout={5000} callback={() => updateState({errors: []})} />
```
rbtying, line 13: huh. Was this really what we were doing before?

I had intended to remove only the entry we just added after 5s (i.e. a new error would not refresh the timeouts on already-displayed errors). Can probably file an issue for this and treat it as nonblocking though!

frontend/src/WebsocketProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+  // https://reactjs.org/docs/hooks-faq.html#why-am-i-seeing-stale-props-or-state-inside-my-function
+  const stateRef = React.useRef(state);
+  const updateStateRef = React.useRef(updateState);
+
+  React.useEffect(() => {
+    stateRef.current = state;
+    updateStateRef.current = updateState;
+  }, [state, updateState]);
+
+  React.useEffect(() => {
```
rbtying, line 37: I'm not quite familiar with what `useEffect` does here -- would we only init the websocket once here? or would we do it every time the props update on `WebsocketProvider`?

frontend/src/WebsocketProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+    );
+    ws.addEventListener('close', () =>
+      updateStateRef.current({connected: false}),
+    );
+    ws.addEventListener('message', (event: MessageEvent) => {
+      const message = JSON.parse(event.data);
+      if (message === 'Kicked') {
+        ws.close();
+      } else {
+        console.log(websocketHandler(stateRef.current, message));
```
rbtying, line 58: remove debug log? (or hide behind debug flag)

frontend/src/websocketHandler.ts
```diff
@@ -0,0 +1,71 @@
...
+  left: WebsocketHandler,
+  right: WebsocketHandler,
+): WebsocketHandler => {
+  return (state: AppState, message: any) => {
+    const newState = {...state, ...left(state, message)};
+    return {...newState, ...right(newState, message)};
+  };
+};
+
+const allHandlers: WebsocketHandler[] = [
```
rbtying, line 64: this is pretty clever!

Curious: do we expect the JS runtime to collapse all of these state-copies into relatively efficient operations? I don't think perf is particularly relevant for this kind of program, but the code as written would imply a linear-time copy of messages/state each time it is updated, right?

rbtying:
aha:
changed the behavior of errors -- instead of removing individual error messages, all disappear at the same time after 5000ms of no new errors

intentional behavior change, I see. Sure, works for me.

frontend/src/Errors.tsx
danwang: no, this is different from what we were doing before. It was hard to get the old behavior working because
- if the responsibility is handled by the webhook handler, the handler can't be just a function
- if the responsibility is handled by the error display, then it needs to be able to hold some state from when it last rendered to prevent new errors from resetting setTimeouts

frontend/src/WebsocketProvider.tsx
danwang: I believe useEffect runs on all renders by default. The second argument to useEffect (here, `[]`) is a list of dependencies that chooses when it should be run. So empty array means that this will only run once, and is equivalent to onComponentMount

frontend/src/WebsocketProvider.tsx
danwang: oops, this was for development debugging. will remove

PR #46, opened 2020-04-11
Extract Play component

danwang:
r? @rbtying 

Extracts the Play component to a separate file, along with some other imports


frontend/src/Play.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const sendEvent = (event: {}) => (send: (value: any) => void) => () => {
+    send(event);
+  };
+  const takeBackCards = sendEvent({Action: 'TakeBackCards'});
+  const endTrick = sendEvent({Action: 'EndTrick'});
+  const startNewGame = sendEvent({Action: 'StartNewGame'});
+
+  const {playPhase} = props;
+
+  // TODO: instead of telling who the player is by checking the name, pass in
```
rbtying, line 39: lol. yes, this is dumb.

rbtying: we can/should just pass the player ID back in the state...

frontend/src/Play.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const canPlay = lastPlay
+    ? selected.length === lastPlay.cards.length
+    : selected.length > 0;
+  const canTakeBack = lastPlay && currentPlayer.id === lastPlay.id;
+
+  const shouldBeBeeping = props.beepOnTurn && isCurrentPlayerTurn;
+
+  const remainingCardsInHands = Object.values(playPhase.hands.hands)
+    .map((playerHand) => {
+      return Object.keys(playerHand).length;
```
rbtying, line 58: `playerHand` is a map<Card, usize>, so we need to sum the inner values too right? (although, when zero this is also zero, so it serves its purpose either way)

danwang: ooooh, I see -- because of the case of duplicate cards. What do you think about using array<Card> format for hands always? I can add some more utils for doing things like add/remove cards from a hand

Will fix in the next commit to this branch

rbtying: the counts data structure is useful on the backend for the actual play validation, and right now the backend datastructure is just serialized to the frontend without modification ( https://github.com/rbtying/shengji/blob/ad156e21d07c369b4912fbe2bb99bd47afa5cd77/core/src/hands.rs#L25 )

I think determining whether the game is over is pretty much the only use case for looking at `hands` right now (they're redacted anyways for other players), since the parallel `cards` array exists to denote ordering

danwang: fixed in 01110a694de54f995383dbc14c87c313706f86d6

danwang:
ptal @rbtying updated in 01110a694de54f995383dbc14c87c313706f86d6 to sum counts of cards

danwang:
yep, I've been going through:
- set up with four players
- deal cards
- play one trick

Issue #32, opened 2020-04-08
improved detail in game log messages in chat


danwang:
<img width="181" alt="Screen Shot 2020-04-07 at 10 21 15 PM" src="https://user-images.githubusercontent.com/935063/78737704-2d0c9300-791e-11ea-93f7-83f3dd3f7886.png">

suggestion: if the backend can return structured data message like
```
{
  type: 'play',
  player: PlayerID,
  cards: Array[Card],
}
```
we could render in the chat more detailed information, like
- rendering cards as displayed number + suit instead of the card text
- making names clickable, or rendering them in a different color depending on team

rbtying:
this seems very reasonable / not very hard to do!

We'd want to update the messages returned from functions on `GameState` to be a new `GameMessage` enum which derives Serialize and Deserialize, and render the relevant strings in the frontend.

PR #48, opened 2020-04-12
Use structured broadcast messages in the game engine

rbtying:
Part 1 of the necessary infrastructure for #32

### DIFF
--- a/frontend/src/Chat.tsx
+++ b/frontend/src/Chat.tsx
@@ -1,7 +1,38 @@
 import * as React from 'react';
 import classNames from 'classnames';
+import {IGameMode} from './types';
 
-export type Message = {from: string; message: string; from_game?: boolean};
+export type MessageVariant =
+  | 'ResettingGame'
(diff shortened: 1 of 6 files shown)


frontend/src/Chat.tsx
```diff
@@ -1,7 +1,38 @@
 import * as React from 'react';
 import classNames from 'classnames';
+import {IGameMode} from './types';
 
-export type Message = {from: string; message: string; from_game?: boolean};
+export type MessageVariant =
+  | 'ResettingGame'
```
danwang, line 6: Are there any ways to get these to serialize to something like `{ResettingGame: true}` instead?

rbtying: there are `serde_json` configuration flags that allow this, but I think it is better to be consistent with the other enums on the wire now, and then change all of them at once.

https://serde.rs/enum-representations.html

internal or external tagging is probably the most JSON-y approach here?

PR #50, opened 2020-04-12
Add layout and fix chat to side

danwang:
r? @rbtying 

Added some layout components and turn the chat into a fixed box on the side. The chat is sticky here, even if scrolling down on the page to see points

Screen Shot 2020-04-12 at 4 01 21 AM

### DIFF
--- a/frontend/src/Chat.tsx
+++ b/frontend/src/Chat.tsx
@@ -35,8 +35,24 @@ const Chat = (props: Props) => {
   };
 
   return (
-    <div className="chat">
-      <div className="messages">
+    <div
+      style={{
+        width: '19%',
+        maxWidth: '300px',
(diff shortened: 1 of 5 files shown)


rbtying:
nice! I'll check this out on a smaller screen to play with it

rbtying:
image
It looks like this might do weird things with the "Reset Game" link (as you change the horizontal size of the window).

As ugly as it might be, I think it would be preferable to reduce the size of the chat box (i.e. less y-height) to allow more space for the cards to be displayed. I'm not very familiar with flex-box, but maybe that's easy?

In practice the chat box is only used to look at game messages, so it doesn't seem like an item of particular focus (in fact, you could imagine having a setting to hide it entirely)

PR #57, opened 2020-04-12
Add inline cards to game log

danwang:
r? @rbtying 

Render "PlayedCard" messages using the data. We can now show inline representations of cards which respects fourcolor mode, e.g.

Screen Shot 2020-04-12 at 5 04 55 PM
Screen Shot 2020-04-12 at 5 05 01 PM

I added a function `convertApiUnion` that turns variants into tagged unions that Typescript can do exhaustive type checking on: e.g.

```
'Tractor' => {type: 'Tractor'}
{Foo: {bar: 'baz'}} => {type: 'Foo', bar: 'baz'}
```


core/src/game_state.rs
```diff
@@ -25,7 +25,7 @@ pub enum MessageVariant {
     NumFriendsSet { num_friends: Option<usize> },
     GameModeSet { game_mode: GameModeSettings },
     TookBackPlay,
-    PlayedCards { cards: Vec<Card> },
+    PlayedCards { player_name: String, cards: Vec<Card> },
```
rbtying: hm. You should have this information from `actor`, right? I don't think it needs to be duplicated here.

frontend/src/InlineCard.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import {AppStateConsumer} from './AppStateProvider';
+import {unicodeToCard, cardToUnicodeSuit, SuitCard} from './util/cardHelpers';
+import ElementWithProps from './ElementWithProps';
+
+const ColoredDiv = (color: string) =>
+  ElementWithProps('span', {
+    style: {color, paddingLeft: '0.1em', paddingRight: '0.1em'},
+  });
+const Black = ColoredDiv('#000000');
```
rbtying, line 10: I have a preference for using the CSS classes (just the unicode suit names) for this because it makes it easier to user CSS -- thoughts?

frontend/src/util/cardHelpers.ts
```diff
@@ -0,0 +1,66 @@
...
+  rank: Rank;
+  suit: Suit;
+};
+
+type Card = SuitCard | {type: 'big_joker'} | {type: 'little_joker'};
+
+// prettier-ignore
+const orderedRanks: (Rank | null)[] = [
+  'A', '2', '3', '4', '5', '6', '7',
+  '8', '9', 'T', 'J', null, 'Q', 'K',
```
rbtying, line 19: I see that you found the "knight" thing in unicode. lol.

Rather than implementing this stuff here, though, can we add additional information as necessary to cards.js, and pass the information from the backend? Having two unicode<->card<->suit mappings feels not ideal.

rbtying:
> I added a function convertApiUnion that turns variants into tagged unions that Typescript can do exhaustive type checking on: e.g.

Can we set serde to generate the tagged union directly instead? it should just be an annotation on `MessageVariant`.

core/src/game_state.rs
danwang: the list of players isn't present in the message. So we'd either need to combine the Message with the GameState, assuming that it is in a state with players, or add it here

frontend/src/util/cardHelpers.ts
danwang: Removed unicode stuff in 638245f6d3464a95ddc717139e42cc100f71a532

danwang:
ptal @rbtying 
- updated serde serialization in f8e8edb5fe7062d86cc7964a3626fee6c55278f7
- responded inline re: `player_name`. Do you have a suggestion for how to access the names? So far, ChatMessage doesn't depend on game_state as a prop

One of benefits of using styles-in-js is that you don't have to deal with the indirect, global nature of CSS sheets. With CSS, the contract/interface is implicit (add a class, and something will happen). We lose static analysis on that -- class names can be dynamically added, and cannot restrict which elements get that class! But if we render all cards through a `InlineCard` or `FullSizedCard`, each component is responsible for its own rendering (including color). Reuse is done like other pieces of code, via exports

frontend/src/util/cardHelpers.ts
danwang: Is there a way to generate cards.js as a file that can be checked in? any code referencing this cannot run in unit tests

rbtying: yeah, we can do that. maybe add a Rust test which asserts the checked in version and the generated version match, too.

danwang:
ptal @rbtying 
Removed inline color choice in 3178c61

here's what it looks like now: 
Screen Shot 2020-04-12 at 8 43 12 PM

PR #60, opened 2020-04-12
Export cards as checked in generated JSON

danwang:
r? @rbtying 

Removes `cards.js` in favor generating a JSON file.

`yarn download-cards-json` downloads and puts it in `src/generated/cards.json`

I didn't write a test to check that they matched

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -20,12 +20,12 @@ use shengji_core::{game_state, interactive, types};
 static NEXT_USER_ID: AtomicUsize = AtomicUsize::new(1);
 
 lazy_static::lazy_static! {
-    static ref CARDS_JS: String = {
+    static ref CARDS_JSON: String = {
         let cards = types::FULL_DECK
             .iter()
             .flat_map(|c| serde_json::to_string(&c.as_info()).ok())
             .collect::<Vec<_>>().join(",");
-        format!("window.CARDS = [{}];", cards)
+        format!("{{\"cards\": [{}]}}", cards)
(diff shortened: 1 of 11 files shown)


backend/src/main.rs
```diff
@@ -20,12 +20,12 @@ use shengji_core::{game_state, interactive, types};
...
 
 lazy_static::lazy_static! {
-    static ref CARDS_JS: String = {
+    static ref CARDS_JSON: String = {
         let cards = types::FULL_DECK
             .iter()
             .flat_map(|c| serde_json::to_string(&c.as_info()).ok())
             .collect::<Vec<_>>().join(",");
-        format!("window.CARDS = [{}];", cards)
+        format!("{{\"cards\": [{}]}}", cards)
```
danwang, line 28: this is necessary because .json files cannot be an array, only an object

backend/src/main.rs
```diff
@@ -475,3 +472,22 @@ static JS_MAP: &str = include_str!("../../frontend/dist/main.js.map");
...
 static WORKER_JS: &str = include_str!("../../frontend/static/timer-worker.js");
+
+#[cfg(test)]
+mod tests {
+    use super::CARDS_JSON;
+
+    static CARDS_JSON_FROM_FILE: &str = include_str!("../../frontend/src/generated/cards.json");
+
+    #[test]
+    fn test_cards_json_compatibility() {
```
danwang, line 483: thanks! I had no idea how to do this 🙈

PR #64, opened 2020-04-13
Use InlineCard for bids

danwang:
r? @rbtying 

Use InlineCard in the bid log message.

Also fix a nil check on `variant.type`

Screen Shot 2020-04-12 at 9 51 20 PM

I changed it to print multiple copies of the card, rather than a number in front because it looked weird when the numbers were the same:
Screen Shot 2020-04-12 at 9 49 49 PM


rbtying:
it might make sense at some point to find a better shorthand for small joker / big joker?

danwang:
sure! at this point we can render anything here (text, an image, some divs, SVG, etc...)  so creativity is the limit :)

PR #91, opened 2020-04-19
Update players/spectators list in Initialize phase

danwang:
r? @rbtying 

Added a new display for users in the Initialize phase. Shows a tabular view rather than a side-by-side view. Buttons for moving/kicking/toggling observer mode are in each row

Screen Shot 2020-04-19 at 1 13 09 AM

Screen Shot 2020-04-19 at 1 15 00 AM


frontend/src/UserList.tsx
```diff
@@ -0,0 +1,133 @@
...
+`;
+
+type RowProps = {
+  player: IPlayer;
+  actions: JSX.Element;
+};
+const UserRow = (props: RowProps) => (
+  <Row key={props.player.name}>
+    <Cell style={{width: '50%'}}>{props.player.name}</Cell>
+    <Cell style={{width: '50%'}}>Rank {props.player.level}</Cell>
```
rbtying, line 58: What do you think about minimizing the space dedicated to the level (and skewing it more towards the player name).

Similarly, we can reduce the size of the overall table and increase readability that way -- the maximum player name length is 32 chars

frontend/src/UserList.tsx
```diff
@@ -0,0 +1,133 @@
...
+    <Cell>{props.actions}</Cell>
+  </Row>
+);
+
+type Props = {
+  currentPlayer: IPlayer;
```
rbtying, line 64: Can we still include the `landlord` information in this component? I don't think players will know to look in the dropdown in continuation settings

frontend/src/UserList.tsx
```diff
@@ -0,0 +1,133 @@
+import * as React from 'react';
+import {IPlayer} from './types';
+import styled from 'styled-components';
+import EmojiButton from './EmojiButton';
+import {WebsocketContext} from './WebsocketProvider';
+
+const Container = styled.div`
+  width: 80%;
```
rbtying, line 8: perhaps, 40% with min-width at 500px or so? On wide screens the table is really wide right now

frontend/src/UserList.tsx
rbtying: This is less of an issue if the overall box is narrower as suggested below

frontend/src/UserList.tsx
danwang: Updated in 554cb908cd44258732970504a2f6c557f3306593

On wide screen, looks like this:

Screen Shot 2020-04-19 at 3 01 04 AM

frontend/src/UserList.tsx
danwang: Updated in fdcce4883c2a56bff0a886edc7aa3ccbbcf251fe

Screen Shot 2020-04-19 at 3 00 22 AM

frontend/src/UserList.tsx
danwang: Changed "Rank" to be minimum width, so players column will eat up all available space

danwang:
ptal @rbtying 

Updated in554cb908cd44258732970504a2f6c557f3306593 and fdcce4883c2a56bff0a886edc7aa3ccbbcf251fe

what if we changed the Play phase player display from left-to-right to something like a circle? A ring could be helpful to see ordering if the player positions don't move from turn to turn. If so, there wouldn't need to be a change from vertical to horizontal

rbtying:
> what if we changed the Play phase player display from left-to-right to something like a circle? A ring could be helpful to see ordering if the player positions don't move from turn to turn. If so, there wouldn't need to be a change from vertical to horizontal

I think we should keep the left-to-right linear view available for a while, since it scales better for large games / large numbers of decks on small screens. Building a ring-variant play UI is probably worthwhile, but it's not something that people have been actively asking for yet.

IMO it would be confusing to have the players in a ring shape if the trick display isn't

PR #92, opened 2020-04-19
Add support for showing card labels on hover (and in settings)

rbtying:
This helps for people who aren't easily able to tell the cards apart.


rbtying:
image

Issue #109, opened 2020-05-08
Allow trump declaration from the kitty


rbtying:
If nobody bids a trump, allow cards from the kitty to be revealed and bid instead.

Two variants:

1. suit of the first card in the kitty becomes the trump
2. suit of the first card of the trump rank in the kitty becomes the trump. If none are found, suit of the highest card in the kitty becomes the trump

Revealing cards from the kitty prevents players from making bids themselves.

Issue #54, opened 2020-04-12
Support alternate bidding strategies


rbtying:
e.g.:

- Bid with your own rank
- Bid with the rank of anyone on the defending team (first person of that rank CCW becomes leader)

PR #111, opened 2020-05-09
Implement bidding from the kitty via revealing cards

rbtying:
Also, implement bidding using the player's current level when the winner of the
last round is unknown.


rbtying:
Fixes #109  and #54

PR #113, opened 2020-05-24
Add statistics tracking across games in localstorage.

katexyu:
Adds functionality to stores game statistics in local storage and adds a component to display statistics.

I added a `GameFinished` message that sends the relevant metadata to the frontend and then the `gameFinishedHandler` figures out what to do to manipulate local storage.

I kind of cargo culted what was going on with the `Settings` for the display part 😇

Screen Shot 2020-05-23 at 8 39 03 PM

Screen Shot 2020-05-23 at 8 15 39 PM

(edge case, no games yet)
Screen Shot 2020-05-23 at 8 29 45 PM

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -925,12 +925,15 @@ impl PlayPhase {
                 });
             }
 
-            game_result.insert(player.name.to_string(), PlayerResult {
-                won_game: landlord_won == self.landlords_team.contains(&player.id),
-                is_defending: is_defending,
-                is_landlord: self.landlord == player.id,
-                ranks_up: num_advances,
-            });
+            game_result.insert(
+                player.name.to_string(),
+                PlayerResult {
+                    won_game: landlord_won == self.landlords_team.contains(&player.id),
+                    is_defending,
+                    is_landlord: self.landlord == player.id,
+                    ranks_up: num_advances,
+                },
+            );
         }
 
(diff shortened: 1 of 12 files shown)


core/src/game_state.rs
```diff
@@ -922,8 +924,22 @@ impl PlayPhase {
                     rank: player.level,
                 });
             }
+
+            game_result.insert(
+                player.name.to_string(),
+                PlayerResult {
+                    won_game: landlord_won == self.landlords_team.contains(&player.id),
```
rbtying, line 931: nit: should this be `landlord_won == `is_defending`?

core/src/types.rs
```diff
@@ -10,6 +10,14 @@ use smallvec::{smallvec, SmallVec};
 #[serde(transparent)]
 pub struct PlayerID(pub usize);
 
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+pub struct PlayerResult {
```
rbtying, line 14: Let's move this to `game_state.rs` and name it `PlayerGameFinishedResult` or something? It's not really a generic game-independent type.

core/src/game_state.rs
katexyu: Ah yeah, I got fixated on the loop to select next landlord but I guess it doesn't have access to `is_defending`

Issue #127, opened 2020-06-07
Simplify friend selection description in pending-friend


cai-jack:
so that instead of 

2 A♦ can be played before the next person to play A♦ is a friend

1 K♦ can be played before the next person to play K♦ is a friend

it will be like:
The person to play 3rd A♦ is a friend

The person to play 2nd K♦ is a friend

Issue #125, opened 2020-06-07
apply the same landlord team background highlight to trick during play


cai-jack:
With 6 or 8 players in a "Finding Friends" game, we find ourselves constantly scrolling up to check for friends in the player list during play to make sure we do not mis-identify friends. Applying the landlord team background highlight with 'landlord' class makes it much easier for a player to identify his/her teammates, especially for "Finding Friends" games in which team mate positions are not constant.

cai-jack:
I already have working code, will check in later after more tests.

PR #128, opened 2020-06-07
#127, simplify friend selection description

cai-jack:
I added changes for #125 in Credits.tsx in this commit as well

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -455,6 +455,12 @@ pub struct Friend {
     player_id: Option<PlayerID>,
 }
 
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+pub struct FriendSelection {
+    card: Card,
+    initial_skip: usize,
+}
+
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum GameState {
@@ -1052,7 +1058,7 @@ impl ExchangePhase {
     pub fn set_friends(
         &mut self,
         id: PlayerID,
-        iter: impl IntoIterator<Item = Friend>,
+        iter: impl IntoIterator<Item = FriendSelection>,
     ) -> Result<(), Error> {
         if self.landlord != id {
             bail!("not the landlord")
@@ -1066,10 +1072,10 @@ impl ExchangePhase {
             if num_friends != friend_set.len() {
                 bail!("incorrect number of friends")
             }
+
+            friends.clear();
+
             for friend in friend_set.iter() {
-                if friend.player_id.is_some() {
-                    bail!("you can't pick your friend on purpose")
-                }
                 if friend.card.is_joker() || friend.card.number() == Some(self.trump.number()) {
                     bail!(
                         "you can't pick a joker or a {} as your friend",
@@ -1079,12 +1085,17 @@ impl ExchangePhase {
                 if self.trump.suit() != None && friend.card.suit() == self.trump.suit() {
                     bail!("you can't pick a trump suit as your friend")
                 }
-                if friend.skip >= self.num_decks {
+                if friend.initial_skip >= self.num_decks {
                     bail!("need to pick a card that exists!")
                 }
+                friends.push(Friend {
+                    card: friend.card,
+                    initial_skip: friend.initial_skip,
+                    skip: friend.initial_skip,
+                    player_id: None,
+                });
             }
-            friends.clear();
-            friends.extend(friend_set);
+
             Ok(())
         } else {
             bail!("not playing finding friends")
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -451,6 +451,7 @@ impl PropagatedState {
 pub struct Friend {
     card: Card,
     skip: usize,
+    initial_skip: usize,
     player_id: Option<PlayerID>,
 }
 
--- a/frontend/src/Exchange.tsx
+++ b/frontend/src/Exchange.tsx
@@ -129,7 +129,6 @@ class Exchange extends React.Component<IExchangeProps, IExchangeState> {
               {this.state.friends.map((friend, idx) => {
                 const onChange = (x: IFriend) => {
                   const new_friends = [...this.state.friends];
-                  x.initial_skip = x.skip;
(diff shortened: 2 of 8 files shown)


rbtying:
ahh, this one is a bit tricky: the skip counter is used in the backend to track the state of the friend-status, so nth isn't quite accurate. (test with picking the 2nd friend, and then watching what happens after the first card is played)

cai-jack:
Got it - Skip counter is updated during play. will use it as the initial value only so that the ordinal number will not change during play. This is one of the more frequent complaints.

rbtying:
I think the best way to do this would be to include another field in the backend for the ordinal number (i.e. modifying this: https://github.com/rbtying/shengji/blob/d126f0b8b5e8dc31ec966ec641f0edcecb3f0834/core/src/game_state.rs#L453)

cai-jack:
@rbtying you are right. That is exactly what I just did. initial_skip was added to store the initial value which does not change during the course of play. Please check it out. Here is a screenshot.

Screen Shot 2020-06-07 at 12 26 30 PM

cai-jack:
also, during exchange state, the "x played in previous hands" is not shown

frontend/src/Exchange.tsx
```diff
@@ -124,10 +125,11 @@ class Exchange extends React.Component<IExchangeProps, IExchangeState> {
           <Trump trump={this.props.state.trump} />
           {this.props.state.game_mode !== 'Tractor' ? (
             <div>
-              <Friends gameMode={this.props.state.game_mode} />
+              <Friends gameMode={this.props.state.game_mode} showPlayed={false} />
               {this.state.friends.map((friend, idx) => {
                 const onChange = (x: IFriend) => {
                   const new_friends = [...this.state.friends];
+                  x.initial_skip = x.skip;
```
rbtying, line 132: Let's only set `initial_skip` in the frontend (i.e. not `x.skip`), and in the backend set `skip = initial_skip`

https://github.com/rbtying/shengji/blob/d126f0b8b5e8dc31ec966ec641f0edcecb3f0834/core/src/game_state.rs#L1068

That way, the backend is protected against someone derping with the websocket (which people do...)

frontend/src/Friends.tsx
```diff
@@ -17,22 +18,12 @@ const Friends = (props: Props) => {
...
-                {friend.skip} <InlineCard card={friend.card} /> can be played
-                before the next person to play <InlineCard card={friend.card} />{' '}
-                is a friend
-              </p>
-            );
-          }
+          return (
+            <p key={idx}>
+              The person to play the {nth(friend.initial_skip + 1)} <InlineCard card={friend.card} /> is a
+              friend. {(props.showPlayed) ? friend.initial_skip - friend.skip + ' played in previous hands' : ''}
```
rbtying, line 24: perhaps, "played in previous tricks"?

frontend/src/Friends.tsx
```diff
@@ -41,4 +32,8 @@ const Friends = (props: Props) => {
   }
 };
 
+
+
+function nth(n: number) { return n + ["st", "nd", "rd"][(((n < 0 ? -n : n) + 90) % 100 - 10) % 10 - 1] || "th" }
```
rbtying, line 37: this is clever!

core/src/game_state.rs
```diff
@@ -451,6 +451,7 @@ impl PropagatedState {
 pub struct Friend {
```
rbtying, line 451: We can define another struct

```
pub struct FriendSelection {
  card: Card,
  initial_skip: usize,
}
```

and use that as an input into `set_friends`. In `set_friends` we can convert them into `Friend` structs.

frontend/src/Exchange.tsx
cai-jack: `friend` is a `&` reference, so the data it refers to cannot be written

frontend/src/Friends.tsx
cai-jack: copied from stack Overflow :)

frontend/src/Exchange.tsx
rbtying: My suggestion would be something like:

```
pub struct FriendSelection {
  card: Card,
  initial_skip: usize,
}

pub fn set_friends(iter: impl IntoIterator<Item = FriendSelection>) -> ... {
  // ...
  let friends = iter.into_iter().map(|fs| Friend {
    card: fs.card,
    initial_skip: fs.initial_skip,
    skip: fs.initial_skip,
    player_id: None,
  });
}

```

frontend/src/Friends.tsx
cai-jack: Sure. I will make it consistent with the current terms

core/src/game_state.rs
cai-jack: need to read more of your code, not clear how to convert from IFriend from wire (frontend)

rbtying: if you derive `Serialize` and `Deserialize`, it should handle the serde for you.

You can look at `GameModeSettings` vs. `GameMode` for a similar pattern.

you'd need to update the corresponding enum in `interactive.rs` that describes an action to use `FriendSelection`, and the function signature for `set_friends`

cai-jack:
@rbtying changes as you requested checked-in. Please verify. Note: I don't know how to use map() to set vec<FriendSelection> into vec<Friend>. I just used a simple iterator. Please let me know if you have a better idea.

frontend/src/Friends.tsx
```diff
@@ -17,22 +18,12 @@ const Friends = (props: Props) => {
...
-                {friend.skip} <InlineCard card={friend.card} /> can be played
-                before the next person to play <InlineCard card={friend.card} />{' '}
-                is a friend
-              </p>
-            );
-          }
+          return (
+            <p key={idx}>
+              The person to play the {nth(friend.initial_skip + 1)} <InlineCard card={friend.card} /> is a
+              friend. {(props.showPlayed) ? friend.initial_skip - friend.skip + ' played in previous tricks' : ''}
```
rbtying, line 24: " were played" rather than "played"?

cai-jack: sentence vs phrase :) don't have to deal with singular/plural cases

Issue #129, opened 2020-06-09
minor issue - skip logging during startup "ERRO Failed to open dump, error: Os { code: 2, kind: NotFound, message: "No such file or directory" }, dump_path: /tmp/shengji_state.json"


cai-jack:
this is not an error during startup since the dump file has not yet been created. Skip logging as an error when it is ErrorKind::NotFound

PR #130, opened 2020-06-09
#129: skip logging NotFound error of game state dump at startup

cai-jack:


backend/src/main.rs
```diff
@@ -173,7 +174,11 @@ async fn main() {
             }
         }
         Err(e) => {
-            error!(init_logger, "Failed to open dump"; "error" => format!("{:?}", e));
+            if let ErrorKind::NotFound = e.kind() {
+                // println!("file not found, not logging");
```
rbtying, line 178: drop the `println!`?

otherwise, sgtm

cai-jack: K. fixed. I could not find a "if not let" to do this. Do u?

rbtying: You can use a match, if you'd like:

```
match e.kind() {
  ErrorKind::NotFound => (),
  _ => {
   // log
  }
}
```

rbtying: Actually, `ErrorKind` implements `Eq`. So you can just do

```
if e.kind() != ErrorKind::NotFound {

}
```

cai-jack: Cool. will do. Thanks

Issue #133, opened 2020-06-12
Consider changing the "landlord" highlight color


rbtying:
The "winner" and "landlord" colors (#efefef and #ddd) are similar, which can be visually confusing when looking at a trick. Let's consider picking a different color for one of them

rbtying:
This became more prevalent after #126 that @cai-jack put together last week

cai-jack:
make all other player list annotations - landlord, you, (next) into trick list as well

PR #136, opened 2020-06-14
#133 - update landlord bg color, and make the trick list carry all th…

cai-jack:
…e info the player list carries

in addition to the bg color change, player list label and coloring for landlord, you, next are all moved down to the trick list.

I spent quite a bit of time trying to get an image file into the code base without success (to be used for showing an icon for the landlord player:)). Don't know enough about the warp crate used in main.rs to do it.

anyway, once the play starts, "you" will have a blue border, can't use 当庄 in the label for playing cards, they are too big. (I wanted to use a bg img to replace it anyway)

winning card will have a bright yellow bg, but only covering the card, not the entire card area, so that team bg color can still be prominent enough.

next player label color is changed to a purplish color so that it displays better than blue.

I played a bit with hiding the player list when play starts, but decided not to do it because observer list still need to be displayed.

Anyway, we probably should look into using some kind of css framework, like bootstrap to get the UI better. Thoughts?

Screen Shot 2020-06-13 at 5 08 23 PM

### DIFF
--- a/frontend/static/style.css
+++ b/frontend/static/style.css
@@ -31,9 +31,10 @@ button { padding: 10px; cursor: pointer; margin: 10px; }
 .hand p { margin: 0; text-align: center; }
 .hand .unselected-cards { padding: 0 10px; }
 .hand .selected-cards { padding: 0 10px; }
-.next { color: #00f }
-.landlord { background: #ddd }
-.landlord .card { background: #ddd }
+.next { color: #b30ba6; }
+.you {border: 3px solid #0575ff;}
+.landlord { background: #4a9a7a }
+.landlord .card { background: #4a9a7a }
 .player { border: 1px #000 solid; padding: 10px; }
 .observer { border: 1px #222 dashed; color: #222; }
 .chat { min-width: 225px; width: 20%; border: 1px #eee solid; float: right; margin-top: 20px; }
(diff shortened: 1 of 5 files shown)


frontend/static/style.css
```diff
@@ -45,9 +46,10 @@ button { padding: 10px; cursor: pointer; margin: 10px; }
 .labeled-play { display: inline-block; padding: 10px; border: 1px solid #000; }
 .labeled-play .label { text-align: center }
 .labeled-play .card { cursor: default; }
-.labeled-play.winning { background: #efefef; }
-.labeled-play.winning .card { background: #efefef; }
-.labeled-play.winning.landlord { background: #ddd; }
+.labeled-play.you {border: 3px solid #0575ff;}
+/* .labeled-play.winning { background: #efed9fa6; } */
```
rbtying, line 50: delete if no longer used?

cai-jack: update as requested. I also removed '.you' from trick since u wanted player list to be on always (I wanted to hide it to have more screen space for showing more useful info like points, and previous trick)

There are a few UI improvements I'd like to see, but without a framework, it may be too much work. For example, the points can be just one line with details hidden, but can be expanded by individual user;  separation of trump suit from non trump suits in hand, etc. Without a css framework, it may be too much work

BTW, reset game should really be restart game since game setting is preserved

cai-jack:
Screen Shot 2020-06-13 at 9 25 03 PM

rbtying:
thank you!

Issue #145, opened 2020-06-20
add save/load/reset game setting functions


cai-jack:
game settings are always set at default when starting up. Adding the functions to persist and load from localStorage will save quite a bit of time and avoid potential mistakes setting up the game.

PR #146, opened 2020-06-20
#145: add functions to preserve and load game settings

cai-jack:


cai-jack:
this is a big change, please review carefully. 

There is a bug in how WebSocket handles timeout: when there are multiple sends pending, only the last timeout is saved. This causes client to disconnect after 5 seconds after game setting is loaded.

To truly fix the big requires adding request sent to have identifiable info added and passed back from server in response. With the current design, it is going to be a huge change.

What I did is to use an array of timer instead of just one to track them, then clear the timeouts as response comes back in. It is an approximation of the fix, but it should work for now. It is simple enough without changing any backend codes.

cai-jack:
Also, lint complains using console.log in tsx files. What is the approved way of logging to console? Thanks.

frontend/src/Initialize.tsx
```diff
@@ -216,6 +235,227 @@ const Initialize = (props: Props) => {
     );
   }
 
+  const [gameModeValue, setGameModeValue] = React.useState(modeAsString);
```
rbtying, line 238: I'm not sure why you need these `useState` values -- wouldn't it suffice to save `props.state.propagated` directly? It should be the closest match to the existing game, rather than any view-local state values

frontend/src/TimerProvider.tsx
```diff
@@ -51,10 +51,13 @@ const _TimerProvider: React.FunctionComponent<{}> = (props) => {
     if (worker !== null) {
       worker.postMessage({command: 'setTimeout', id, timeout: delay});
     }
+    // console.log("TimerProvider set timeout: " + id);
```
rbtying, line 54: we shouldn't commit console logs to production code -- it tends to make everything run slower, and we shouldn't debug in production anyways.

frontend/src/Initialize.tsx
cai-jack: I tried that, it did not work. I set the values in the select list first, hoping that it would trigger the change event, it did not. So I did what I did, kind of a pain.

I guess I can send the change config requests directly to the server first, then the response will cause the client to re-render its value without all these ugly codes, like resetGameSettings does! Will fix this.

cai-jack:
The timeout of send issue took a lot of time to figure out!
It happens when the load config sends a list of change config commands to the server. Since it goes through the list so quickly, no response has been received for the previous send before the next send arrives. the send code only keeps track the current timer id, overwriting any timer that has not been cleared. When a response eventually arrives, the message handler can only obtain the latest timer id to clear the timeout even though there are other timers still running, but their ids are now lost. When those untracked timers expire, they set the connected to false, the client then disconnect. It is easy to reproduce, you just need to revert my changes in WebSocket, and click either the load or the reset button. You will get a timeout in 5 seconds

rbtying:
We can check whether the current timerRef matches the one that we passed into the callback before calling setState, which should fix this with substantially less complexity

cai-jack:
I thought about using one new API to do this and decided not to pursue that route. the individual message on setting changes  clients received are still desirable, and there is only one place/one way to change setting. Using a separate API will make maintenance harder when new configs are added, for example. Also, it is unlikely in the current setting, the config will be partially changed. If it is, it can be fixed by simply clicking load again:)

cai-jack:
I don't see how we can check that reliably. The load function sends about 10 API requests, none was responded until all of the sends are completed, the send() creates 10 timers, only the last one is tracked, the other 9 are lost. 

my current fix (more like a work-around) assume this usage pattern, and clears the timeouts sequentially. The ultimate fix will have to be sending the timer id along each request to server, and have the server piggy back in the response header to have a solid fix.

frontend/src/TimerProvider.tsx
cai-jack: got it. Just makes debugging a bit harder:)

cai-jack:
prettier? It did not complain in the ci run. I will look into it. 

How do you run prettier?

cai-jack:
from command line

rbtying:
> prettier? It did not complain in the ci run. I will look into it.
> 
> How do you run prettier?

I use `yarn prettier --write` with

```
$ prettier --version
2.0.2
```

rbtying:
```diff --git a/frontend/src/WebsocketProvider.tsx b/frontend/src/WebsocketProvider.tsx
index 14abef6..311e789 100644
--- a/frontend/src/WebsocketProvider.tsx
+++ b/frontend/src/WebsocketProvider.tsx
@@ -90,11 +90,14 @@ const WebsocketProvider: React.FunctionComponent<{}> = (props) => {
     }
     // We expect a response back from the server within 5 seconds. Otherwise,
     // we should assume we have lost our websocket connection.
-    setTimerRef.current(
-      setTimeoutRef.current(() => {
+
+    const localTimerRef = setTimeoutRef.current(() => {
+      if (timerRef.current === localTimerRef) {
         updateStateRef.current({connected: false});
-      }, 5000),
-    );
+      }
+    }, 5000);
+
+    setTimerRef.current(localTimerRef);
     websocket?.send(JSON.stringify(value));
   };
   // TODO(read this from consumers instead of globals)```

this seems sufficient to resolve the issue?

**Comment by cai-jack** [comment]
here is mine: 
yarn prettier --version
yarn run v1.22.4
$ prettier src --version
2.0.2


I never ran manually, only through vscode's format command, btw

**Comment by cai-jack** [comment]
Not likely. It still keeps track of only one timer if I understand your change correctly. The bug is when there are more than one send pending without response from the server, the next send will overwrite tracking of the timer id(s) of the previous send(s), these lost tracked ids still has a corresponding timer running, respectively. the message handler just has no way to clearing it out. It is very clear with console.log how the sequence of send/message affects clearing of timers.

Anyway, I will give it a try to see if it can handle the load function.

**Comment by rbtying** [comment]
> Not likely. ...
from the concurrency point of view, the condition we care about isn't the number of timers or what timer is firing, but instead whether the state of the world at send-time has changed since the callback was scheduled. If it has changed, it's either because we got a message or we called `send` again. If it hasn't, then we probably lost our connection to the server.

**Comment by rbtying** [comment]
BTW: I stashed the race-condition fix in #147 and the prettier configuration (and adding it to CI) in #148). So, once I merge both of those, this change should be simpler, I hope?

It would be ideal not to do any frontend state management for variables that are also set in the backend, to avoid data-consistency issues

File: frontend/src/Initialize.tsx [hunk 3]
```diff
@@ -216,6 +217,188 @@ const Initialize = (props: Props) => {
...
+                SetThrowEvaluationPolicy: value,
+              },
+            });
+            break;
+        }
+      }
+    }
+  }
+
+  const resetGameSettings = (evt: any) => {
```
rbtying, line 329: Hm. Rather than having a `resetGameSettings` which requires us to hardcode all of the default values, what do you think adding an endpoint that returns a default `PropagatedState`? e.g. at /default_settings.

We could then XHR request it, and the code you've already written for loading game settings from a serialized propagated state.

cai-jack:
how do I run yarn prettier? thanks. vscode format did not help. Thanks.

frontend/src/Initialize.tsx
cai-jack: sounds good since you already have the default settings in the backend

cai-jack:
> > Not likely. ...
> 
> from the concurrency point of view, the condition we care about isn't the number of timers or what timer is firing, but instead whether the state of the world at send-time has changed since the callback was scheduled. If it has changed, it's either because we got a message or we called `send` again. If it hasn't, then we probably lost our connection to the server.

I see your change now:) Instead of setting up a timeout for every pending send and track them, you skip setting up the timeout! This will work if the request does not actually time out at all, kind of defeating the purpose of timeout:). Anyway, if this is what you prefer, I am fine with it. Chance of this happens is very low.

rbtying:
> Chance of this happens is very low.

If it helps, this was also the previous behavior, and is intentional: the check exists because some browsers (e.g. Chrome) don't notify the user that a websocket connection has closed under certain network-change conditions (e.g. changing WiFi, changing access point, etc), and so all messages end up getting blackholed. In practice, the current server implementation is also fully synchronized and can only service a single request at a time (though it does so very quickly), so a single request taking too long will head-of-line blocking

frontend/src/Initialize.tsx
```diff
@@ -219,22 +219,25 @@ const Initialize = (props: Props) => {
...
     evt.preventDefault();
-    const settings = localStorage.getItem('gameSettingsInLocalStorage');
+    const settings = localStorage.getItem("gameSettingsInLocalStorage");
     if (settings !== null) {
       const gameSettings: IPropagatedState = JSON.parse(settings);
       for (const [key, value] of Object.entries(gameSettings)) {
         switch (key) {
-          case 'game_mode':
-            if (value === 'Tractor') {
+          case "game_mode":
```
rbtying, line 235: I think you can actually just do

```
send({Action:{SetGameMode: value}})
```
similar to the other cases? the format should be compatible on the wire.
```

File: frontend/src/Initialize.tsx [hunk 5]
```diff
@@ -219,22 +219,25 @@ const Initialize = (props: Props) => {
 
   const saveGameSettings = (evt: any) => {
     evt.preventDefault();
-    localStorage.setItem('gameSettingsInLocalStorage', JSON.stringify(props.state.propagated));
+    localStorage.setItem(
```
rbtying, line 222: we might want to use the abstractions here: https://github.com/rbtying/shengji/blob/08b6f8c89213ce4b11def917b6f4f191a3f060c2/frontend/src/localStorageState.ts#L26 

though I think in practice it's basically the same?

frontend/src/Initialize.tsx
```diff
@@ -219,22 +219,25 @@ const Initialize = (props: Props) => {
...
+      JSON.stringify(props.state.propagated)
+    );
   };
 
   const loadGameSettings = (evt: any) => {
     evt.preventDefault();
-    const settings = localStorage.getItem('gameSettingsInLocalStorage');
+    const settings = localStorage.getItem("gameSettingsInLocalStorage");
     if (settings !== null) {
       const gameSettings: IPropagatedState = JSON.parse(settings);
```
rbtying, line 232: we should also handle the case where JSON parsing fails, and not crash the UI?

frontend/src/Initialize.tsx
cai-jack: why does persist has a before param that is never used?

this looks strange:

    stringLocalStorageState("gameSettingsInLocalStorage", '').persist(null, JSON.stringify(props.state.propagated));

  much simpler using localStorage directly.

cai-jack: also, to use the instance generated by stringLocalStorageState for retrieval, a const needs to be created every time Initialize is called, not that efficient. I will stick with using localStorage directly for now.

rbtying: hm. the intent with the abstraction is to keep things type-safe -- the `stringLocalStorageState` wraps a serializer and deserializer from the underlying localstorage, which you are using JSON.stringify / etc for.

I can refactor this later if it's confusing -- not a big deal for now.

cai-jack:
I think it will be very helpful to allow console.log when there is an error in the front end. It will help quite a bit in diagnostics

cai-jack:
something is wrong, reset does not work after I merge your changes. Please hold off merge

backend/src/main.rs
```diff
@@ -332,6 +341,12 @@ async fn get_stats(
     }))
 }
 
+async fn get_default_game_settings(
+    setting: game_state::PropagatedState,
+) -> Result<impl warp::Reply, warp::Rejection> {
+    Ok(warp::reply::json(&setting))
```
rbtying, line 347: I think you can just have

```
Ok(warp::reply::json(&game_state::PropagatedState::default()))
```

and not take `setting` as an argument -- might make it simpler?

cai-jack: oh, I tried that, but don't know how to skip .and().

rbtying:
```diff --git a/backend/src/main.rs b/backend/src/main.rs
index 6225a12..3c17a26 100644
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -243,6 +243,7 @@ async fn main() {
     let game_stats = warp::path("stats")
         .and(games)
         .and_then(|(game, stats)| get_stats(game, stats));
+    let default_settings = warp::path("default_settings.json").and_then(default_propagated);
     let routes = index
         .or(js)
         .or(js_map)
@@ -252,11 +253,16 @@ async fn main() {
         .or(api)
         .or(rules)
         .or(dump_state)
-        .or(game_stats);
+        .or(game_stats)
+        .or(default_settings);

     warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
 }

+async fn default_propagated() -> Result<impl warp::Reply, warp::Rejection> {
+    Ok(warp::reply::json(&game_state::PropagatedState::default()))
+}
+
 async fn dump_state(games: Games) -> Result<impl warp::Reply, warp::Rejection> {
     let mut state_dump: HashMap<String, game_state::GameState> = HashMap::new();
     let mut games = games.lock().await;```

should work

Comment (cai-jack) [reply backend/src/main.rs:347]: Thanks. You must have good hands! I did the same, and vscode complained! It is probably a delayed validation from vscode

**Comment by cai-jack** [comment]
> something is wrong, reset does not work after I merge your changes. Please hold off merge

fixed now. I added another json parsing when result is already parsed for some reason

File: frontend/src/Initialize.tsx [hunk 8]
```diff
@@ -216,6 +217,146 @@ const Initialize = (props: Props) => {
...
+  const loadGameSettings = (evt: any) => {
+    evt.preventDefault();
+    const settings = localStorage.getItem("gameSettingsInLocalStorage");
+    if (settings !== null) {
+      let gameSettings: IPropagatedState;
+      try {
+        gameSettings = JSON.parse(settings);
+        setGameSettings(gameSettings);
+      } catch (err) {
+        localStorage.setItem(
```
rbtying, line 338: in the error case, we probably want to unset the value in localstorage (rather than setting it to the current value)?

frontend/src/Initialize.tsx
```diff
@@ -216,6 +217,146 @@ const Initialize = (props: Props) => {
...
+            });
+            break;
+          case "num_decks":
+            send({
+              Action: {
+                SetNumDecks: value,
+              },
+            });
+            if (kittySizeSet) {
+              // reset the size again, as setting deck numn resets kitty_size to default
```
rbtying, line 248: nit: number or num, not numn

frontend/src/Initialize.tsx
```diff
@@ -216,6 +217,146 @@ const Initialize = (props: Props) => {
...
+          JSON.stringify(props.state.propagated)
+        );
+      }
+    }
+  };
+
+  const resetGameSettings = (evt: any) => {
+    evt.preventDefault();
+
+    fetch("/default_settings")
```
rbtying, line 349: 1. any chance we can rename `default_settings` to `default_settings.json`?
2. should we be using a relative path here? the deployed URL is in a subdirectory

rbtying:
other than serde-default, this seems fine. Next time, I would appreciate it if you split the separate feature into its own PR.

Probably want to invest in making the "initialize game" thing a bit more organized sometime soon; the number of settings is really stacking up.

frontend/src/Trick.tsx
```diff
@@ -95,7 +96,9 @@ const Trick = (props: IProps) => {
             key={id}
             id={id}
             label={
-              namesById[id] + (id === props.landlord ? " (当庄)" : "") + suffix
+              namesById[id] +
```
rbtying, line 99: if `landlord_emoji` isn't set, we should still show (当庄) ideally?

cai-jack: That is exactly what it is. I want to keep the default as (当庄) to make you happen :)

cai-jack:
> Next time, I would appreciate it if you split the separate feature into its own PR.

will do. I lump it together since I wanted to test the save/load/reset function as well
> 
> Probably want to invest in making the "initialize game" thing a bit more organized sometime soon; the number of settings is really stacking up.

I have more options in mind:) one of the reasons I asked about use bootstrap, actually

frontend/src/Trick.tsx
rbtying: oh -- is `landlord_emoji` really just `landlord_label` then? I was confused

rbtying: > > Probably want to invest in making the "initialize game" thing a bit more organized sometime soon; the number of settings is really stacking up.
> 
> I have more options in mind:) one of the reasons I asked about use bootstrap, actually

doing some more UI work here seems reasonable -- not sure about Bootstrap in particular since we really only have multiselects here. You could imagine using `react-select` for this.

I think the more important bit is going to be figuring out which of these settings can live in a settings pane vs. being available on the front page, since making configuration complicated is less good for players

cai-jack: > oh -- is `landlord_emoji` really just `landlord_label` then? I was confused

no. The landlord label has player name as well. this emoji string is the tail part of the label, so the landlord label is <player name>' '(当庄) or <player name>' '<emoji>. the emoji string is just a utf8 string

cai-jack: > > oh -- is `landlord_emoji` really just `landlord_label` then? I was confused
> 
> no. The landlord label has player name as well. this emoji string is the tail part of the label, so the landlord label is ' '(当庄) or ' '. the emoji string is just a utf8 string

so the landlord label is <<player_name>>' '(当庄) or <<player_name>>' '<<emoji>>. the emoji string is just a utf8 string

cai-jack: or <<player_name>>' '<<emoji>>

cai-jack: stupid! or <<player_name>>' '<<landlord_emoji>>

frontend/src/Play.tsx
```diff
@@ -91,7 +95,7 @@ const Play = (props: Props) => {
         trick={playPhase.trick}
         players={playPhase.propagated.players}
         landlord={playPhase.landlord}
-        landlord_emoji={playPhase.propagated.landlord_emoji}
+        landlord_emoji={emoji}
```
rbtying, line 98: let's name this `landlord_suffix` then? since it's not always an emoji

frontend/src/Initialize.tsx
rbtying: I think this still needs to be updated :)

cai-jack: Sure. I was only following your convention of stats not stats.json :)

Issue #156, opened 2020-06-25
option to select landlord on first bid instead of winning bid


cai-jack:
FirstLandlordSelectioPolicy has two values, the default is ByWinningBid, which is the current policy; a new value BidFirstBid is added so that when no landlord has been chosen yet (first time starting the game), the landlord is decided by the first bid, not the winning bid, while trump is decided by winning bid.

PR #157, opened 2020-06-25
#156: add FirstLandlordSelectionPolicy

cai-jack:
see the two screen shots

pic1: test bid first, the other two bid later, with diamond 2 winning the bid, only test can pick up
pic2: after pickup, test is the landlord, and trump is diamond 2


cai-jack:
Screen Shot 2020-06-24 at 6 33 17 PM
Screen Shot 2020-06-24 at 6 33 33 PM

core/src/game_state.rs
```diff
@@ -1480,7 +1504,20 @@ impl DrawPhase {
             bail!("nobody has bid yet")
         } else {
             let winning_bid = bail_unwrap!(self.autobid.or_else(|| self.bids.last().copied()));
-            let landlord = self.propagated.landlord.unwrap_or(winning_bid.id);
+            let first_bid = bail_unwrap!(self.autobid.or_else(|| self.bids.first().copied()));
+            let landlord;
+
+            if self.propagated.first_landlord_selection_policy
```
rbtying, line 1510: Perhaps,

```
landlord = match self.propagated.first_landlord_selection_policy {
  FirstLandlordSelectionPolicy::ByWinningBid => self.propagated.landlord.unwrap_or(winning_bid.id),
  FirstLandlordSelectioNPolicy::ByFirstBid => self.propagated.landlord.unwrap_or(first_bid.id),
};
```

?

cai-jack: let landlord = match self.propagated.first_landlord_selection_policy {
                FirstLandlordSelectionPolicy::ByWinningBid => {
                    self.propagated.landlord.unwrap_or(winning_bid.id)
                }
                FirstLandlordSelectionPolicy::ByFirstBid => {
                    if self.propagated.landlord.is_none() {
                        self.propagated.landlord.unwrap_or(first_bid.id)
                    } else {
                        self.propagated.landlord.unwrap_or(winning_bid.id)
                    }
                }
            };

core/src/game_state.rs
```diff
@@ -1480,7 +1504,20 @@ impl DrawPhase {
...
         } else {
             let winning_bid = bail_unwrap!(self.autobid.or_else(|| self.bids.last().copied()));
-            let landlord = self.propagated.landlord.unwrap_or(winning_bid.id);
+            let first_bid = bail_unwrap!(self.autobid.or_else(|| self.bids.first().copied()));
+            let landlord = match self.propagated.first_landlord_selection_policy {
+                FirstLandlordSelectionPolicy::ByWinningBid => {
+                    self.propagated.landlord.unwrap_or(winning_bid.id)
+                }
+                FirstLandlordSelectionPolicy::ByFirstBid => {
+                    if self.propagated.landlord.is_none() {
```
rbtying, line 1513: I don’t think you need this if statement— unwrap_or only uses the right hand side when the left hand side is None

cai-jack:
btw, how did you manage to remove desc.d.ts and still make it work? @rbtying  Thanks.

rbtying:
Heh, I didn’t really — something in the package configuration is messed up so that the d.ts file in node modules isn’t getting picked up. The current hack I have is just to copy the file in to the source list, so that at least there are types

cai-jack:
yeah, just saw your changes. Thanks.

Issue #160, opened 2020-06-26
game option to restrict joker bid length


cai-jack:
currently when the current high bid length is 2 or higher, joker bids can overbid with equal or higher length. This option enables setting joker bids to overbid with higher length only

PR #161, opened 2020-06-26
#160: add joker overbid game option; minor UI improvement on landlord…

cai-jack:
… emoji selection

"higher length" or "longer length"?


cai-jack:
Screen Shot 2020-06-26 at 9 49 59 AM

core/src/game_state.rs
```diff
@@ -1452,10 +1476,26 @@ impl DrawPhase {
                         } else if new_bid.count == existing_bid.count {
                             match (new_bid.card, existing_bid.card) {
                                 (Card::BigJoker, Card::BigJoker) => (),
-                                (Card::BigJoker, _) => valid_bids.push(new_bid),
+                                (Card::BigJoker, _) => {
```
rbtying, line 1479: you can do

```
(Card::BigJoker, _) if self.propagated.joker_overbid_selection_policy == JokerOverBidSelectionPolicy::AllowEqualOrGreaterLength => {

}
```

to reduce the nesting depth here
```

File: core/src/interactive.rs [hunk 2]
```diff
@@ -345,6 +354,8 @@ impl BroadcastMessage {
             FriendSelectionPolicySet { policy: FriendSelectionPolicy::HighestCardNotAllowed} => format!("{} disallowed the highest non-trump card, as well as trump cards, from being selected as a friend", n?),
             FirstLandlordSelectionPolicySet { policy: FirstLandlordSelectionPolicy::ByWinningBid} => format!("{} set winning bid to decide both landlord and trump", n?),
             FirstLandlordSelectionPolicySet { policy: FirstLandlordSelectionPolicy::ByFirstBid} => format!("{} set first bid to decide landlord, winning bid to decide trump", n?),
+            JokerOverbidSelectionPolicySet { policy: JokerOverbidSelectionPolicy::AllowEqualOrGreaterLength} => format!("{} set joker overid can be of the same or greater suit length of current bid when current bid length is 2 or higher", n?),
```
rbtying, line 357: "{} allowed joker bids to outbid non-joker bids with the same number of cards"?

core/src/interactive.rs
```diff
@@ -345,6 +354,8 @@ impl BroadcastMessage {
...
+            JokerOverbidSelectionPolicySet { policy: JokerOverbidSelectionPolicy::AllowOnlyGreaterLength} => format!("{} set Joker overid can be of only greater suit length of current bid", n?),
```
rbtying, line 358: "{} requires all bids to have more cards than the previous bids"

core/src/game_state.rs
```diff
@@ -158,6 +158,18 @@ impl Default for FirstLandlordSelectionPolicy {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum JokerOverbidSelectionPolicy {
```
rbtying, line 162: perhaps just `BidPolicy`, and the options are `JokerOrGreaterLength` and `GreaterLength`?

cai-jack:
how do you run clippy on your own codes only? Thanks.

rbtying:
```
cargo clippy
```
should work. If you've already built the code, it will noop, so you might need
```
cargo clean -p ${crate_name} && cargo clippy
```

core/src/game_state.rs
cai-jack: added by cargo fmt, ci will fail without the extra brackets

cai-jack:
got it. thanks.

core/src/game_state.rs
rbtying: did this get changed ?

cai-jack:
yarn --fix is changed to yarn --write now. Readme should be updated. @rbtying

Issue #163, opened 2020-06-28
game option to reward bonus level to landlord team with smaller than normal size


cai-jack:
for successfully defend its game

PR #164, opened 2020-06-28
#163: add game option for reward bonus level to landlord team with le…

cai-jack:
…ss than normal size.

so in a 6 or 7 player game, if the landlord team is only 2 players, each player gets a bonus level if they won the game, for example

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -170,6 +170,18 @@ impl Default for BidPolicy {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum BonusLevelPolicy {
+    NoBonusLevel,
+    BonusLevelForSmallerLandlordTeam,
+}
+
+impl Default for BonusLevelPolicy {
+    fn default() -> Self {
+        BonusLevelPolicy::NoBonusLevel
+    }
+}
+
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct PropagatedState {
     pub game_mode: GameModeSettings,
@@ -206,6 +218,8 @@ pub struct PropagatedState {
     first_landlord_selection_policy: FirstLandlordSelectionPolicy,
     #[serde(default)]
     bid_policy: BidPolicy,
+    #[serde(default)]
+    bonus_level_policy: BonusLevelPolicy,
 }
 
 impl PropagatedState {
@@ -492,6 +506,18 @@ impl PropagatedState {
         if policy != self.advancement_policy {
             self.advancement_policy = policy;
             Ok(vec![MessageVariant::AdvancementPolicySet { policy }])
+        } else {
+            Ok(vec![])
+        }
+    }
+
+    pub fn set_bonus_level_policy(
+        &mut self,
+        policy: BonusLevelPolicy,
+    ) -> Result<Vec<MessageVariant>, Error> {
+        if policy != self.bonus_level_policy {
+            self.bonus_level_policy = policy;
+            Ok(vec![MessageVariant::BonusLevelPolicySet { policy }])
         } else {
             Ok(vec![])
         }
@@ -1061,7 +1087,12 @@ impl PlayPhase {
         }
 
         let (non_landlord_level_bump, landlord_level_bump, landlord_won) =
-            Self::compute_level_deltas(self.num_decks, non_landlords_points);
+            Self::compute_level_deltas(
+                self.num_decks,
+                non_landlords_points,
+                self.propagated.bonus_level_policy,
+                self.landlords_team.len() < self.propagated.players.len() / 2,
+            );
 
         let mut propagated = self.propagated.clone();
 
@@ -1109,20 +1140,32 @@ impl PlayPhase {
     pub fn compute_level_deltas(
         num_decks: usize,
         non_landlords_points: isize,
+        bonus_level_policy: BonusLevelPolicy,
+        smaller_landlord_team_size: bool,
     ) -> (usize, usize, bool) {
         let point_segments = (num_decks * 20) as isize;
         let landlord_won = non_landlords_points < 2 * point_segments;
+        let bonus_level;
+
+        if landlord_won
+            && bonus_level_policy == BonusLevelPolicy::BonusLevelForSmallerLandlordTeam
+            && smaller_landlord_team_size
+        {
+            bonus_level = 1;
+        } else {
+            bonus_level = 0;
+        };
 
         if landlord_won && non_landlords_points <= 0 {
             (
                 0,
-                (3 - non_landlords_points / point_segments) as usize,
+                bonus_level + ((3 - non_landlords_points / point_segments) as usize),
                 true,
             )
         } else if landlord_won {
             (
                 0,
-                (2 - non_landlords_points / point_segments) as usize,
+                bonus_level + ((2 - non_landlords_points / point_segments) as usize),
                 true,
             )
         } else {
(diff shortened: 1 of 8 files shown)


core/src/game_state.rs
```diff
@@ -170,6 +170,18 @@ impl Default for BidPolicy {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum BonusLevelPolicy {
```
rbtying, line 174: I don't actually think this needs to be optional; I just never got around to implementing it. Might make sense to just change the default behavior here.

core/src/game_state.rs
```diff
@@ -1061,7 +1087,12 @@ impl PlayPhase {
         }
 
         let (non_landlord_level_bump, landlord_level_bump, landlord_won) =
-            Self::compute_level_deltas(self.num_decks, non_landlords_points);
+            Self::compute_level_deltas(
+                self.num_decks,
+                non_landlords_points,
+                self.propagated.bonus_level_policy,
+                self.landlords_team.len() < self.propagated.players.len() / 2,
```
rbtying, line 1094: there is an explicit (overridable) setting for `num_friends`, so we should use that instead of recomputing it here -- the rule is a bonus for "smaller than expected" team, not "smaller than default"

core/src/game_state.rs
cai-jack: let's keep it an option if it is OK with you as there are people think that landlord team gets into this situation mostly by being careless, and should not deserve a bonus :)

I will change the default to bonus, however

core/src/game_state.rs
cai-jack: good catch. Will fix. Thanks.

core/src/game_state.rs
```diff
@@ -1060,8 +1086,40 @@ impl PlayPhase {
             }
         }
 
+        let mut smaller_landlord_team = false;
```
rbtying, line 1089: I think you actually want

```
let smaller_landlord_team = match self.game_mode {
  GameMode::FindingFriends { num_friends, ref friends } if num_friends > friends.flat_map(|f| f.player_id).collect<HashSet<PlayerID>>().len() => true,
  _ => false,
};
```

In order for this to be correct -- `num_friends` and `friends.len()` are the same, and the interesting bit for how many players are _actually_ on the team is how many unique `friend.player_id`

cai-jack: not sure this is correct. I did try something similar to this, and had to back out the changes. The game_mode from the state has the actual current number of friends in num_friends and friends holds the current list of friends, they actually match. the game_mode of the propagated has the setting. I checked using the debugger to confirm as well. Please let me know if I missed anything. Will update the UI.

rbtying: https://github.com/rbtying/shengji/blob/9bb25ba240d903901aa2db82e3bb29e8f3133f32/core/src/game_state.rs#L1683

^ This is where `num_friends` is taken from `propagated` to start the game -- if you set "num_friends" too high, it might have broken your experiment? (in that case, we would correctly want to use the value in `self.game_mode.num_friends`

I realized, actually, that it's insufficient to just check the number of unique player IDs < num_friends, but also that the landlord's ID should not be one of those player IDs. So the comparison needs to account for that as well

If you write a test for this case, I can poke at why it is doing something unexpected?

cai-jack: Here how I checked. 
1. played with 4 player
2. landlord select a friend
3. during play, landlord picked himself as friend
4. landlord won the game
5. when I used only state.game_mode, bonus level was not added. I used the debugger to check, num_friends is 1, and friends.len is one as well. 
ah, as I am writing this,  you are right num_friends is the configured number of friends (EXCLUDING the landlord), friends.len is the actual friends list, including the landlord. This should simplify the test!

cai-jack: num_friends can never be too high, it is capped at players.len()/2

cai-jack: players.len()/2 - 1

frontend/src/Points.tsx
```diff
@@ -71,13 +72,29 @@ const Points = (props: IProps): JSX.Element => {
...
+      thresholdStr = `${landlord.name}'s team will go up 2 levels (next threshold: ${segment}分)`;
+    }
   } else if (nonLandlordPointsWithPenalties < 2 * segment) {
-    thresholdStr = `${
-      landlord.name
-    }'s team will go up 1 level (next threshold: ${2 * segment}分)`;
+    if (props.bonusLevel) {
+      thresholdStr = `${
+        landlord.name
+      }'s team will go up 2 level with bonus level (next threshold: ${
```
rbtying, line 90: 2 levels, including a small-team bonus

frontend/src/Points.tsx
```diff
@@ -71,13 +72,29 @@ const Points = (props: IProps): JSX.Element => {
   let thresholdStr = "";
 
   if (nonLandlordPointsWithPenalties === 0) {
-    thresholdStr = `${landlord.name}'s team will go up 3 levels (next threshold: 5分)`;
+    if (props.bonusLevel) {
+      thresholdStr = `${landlord.name}'s team will go up 4 levels with bonus level (next threshold: 5分)`;
```
rbtying, line 76: 4 levels, including a small-team bonus

core/src/game_state.rs
```diff
@@ -1060,8 +1086,32 @@ impl PlayPhase {
...
+        }
+
         let (non_landlord_level_bump, landlord_level_bump, landlord_won) =
-            Self::compute_level_deltas(self.num_decks, non_landlords_points);
+            Self::compute_level_deltas(
+                self.num_decks,
+                non_landlords_points,
+                self.propagated.bonus_level_policy,
+                smaller_landlord_team,
+            );
```
rbtying, line 1114: might make sense to emit a "bonus level" message if one is applied, to keep things from being surprising ?

cai-jack: hang on:) I am generating a giant test case to play thru a 6 player game. I think I find the simplest test. The test in points.tsx is the actually the most efficient one. I am using it in game_state.rs as well

        let mut smaller_landlord_team = false;

        if let GameMode::FindingFriends {
            num_friends,
            friends: _,
        } = &self.game_mode
        {
            let actual_team_size: usize;
            let setting_team_size = *num_friends + 1;

            actual_team_size = self.landlords_team.len();

            if actual_team_size < setting_team_size {
                smaller_landlord_team = true;
            } else {
                smaller_landlord_team = false;
            }
        }

when I ran the debugger for testing, frends.len is actually the same as num_friends, its content filled with friendSelection element. So this one should be the best solution.

rbtying: hm, I don't recall if `self.landlords_team` is uniqued in the case of the same player joining the team twice -- would need to check that for sure. 

nit: `smaller_landlord_team = actual_team_size < setting_time_size` is more concise and says the same thing... and you can use an `else { false }` for the if statement to have an immutable `smaller_landlord_team`

cai-jack: hmm, I can't find anywhere landlords_team can contain duplicate players, in fact, in a few places, assertion was added to ensure it was size 1 when first player was added. Can you confirm? Thanks.

game.users does contain duplicate names

cai-jack: I can confirm landlords_team does not contain duplicates by joining the game with duplicate landlord name.

rbtying: It looks like the code does currently check if the player is already in `landlords_team` prior to adding them, so it is already deduplicated!

I think this test covers an additional case (same non-landlord player joining twice), and is much shorter:
```
    #[test]
    fn test_landlord_small_team() {
        let mut init = InitializePhase::new();
        init.set_game_mode(GameModeSettings::FindingFriends {
            num_friends: Some(3),
        })
        .unwrap();
        let p1 = init.add_player("p1".into()).unwrap().0;
        let p2 = init.add_player("p2".into()).unwrap().0;
        let p3 = init.add_player("p3".into()).unwrap().0;
        let p4 = init.add_player("p4".into()).unwrap().0;
        let p5 = init.add_player("p5".into()).unwrap().0;
        let p6 = init.add_player("p6".into()).unwrap().0;
        let p7 = init.add_player("p7".into()).unwrap().0;
        let p8 = init.add_player("p8".into()).unwrap().0;

        init.set_landlord(Some(p1)).unwrap();
        init.set_rank(p1, Number::Seven).unwrap();

        let mut draw = init.start().unwrap();
        let mut deck = vec![];

        // We need at least two cards per person, since the landlord needs to
        // bid, and the biddable card can't be the friend-selection card.
        let p1_hand = vec![cards::S_7, cards::D_3];
        let p2_hand = vec![cards::D_4, cards::D_5];
        let p3_hand = vec![cards::C_6, cards::C_8];
        let p4_hand = vec![cards::C_9, cards::C_10];
        let p5_hand = vec![cards::C_J, cards::C_Q];
        let p6_hand = vec![cards::C_K, cards::C_A];
        let p7_hand = vec![cards::H_2, cards::H_3];
        let p8_hand = vec![cards::H_4, cards::H_5];

        // Set up the deck to have the appropriate cards.
        for i in 0..2 {
            deck.push(p1_hand[i]);
            deck.push(p2_hand[i]);
            deck.push(p3_hand[i]);
            deck.push(p4_hand[i]);
            deck.push(p5_hand[i]);
            deck.push(p6_hand[i]);
            deck.push(p7_hand[i]);
            deck.push(p8_hand[i]);
        }
        deck.reverse();
        draw.deck = deck;
        draw.position = 0;

        // Draw the deck
        for _ in 0..2 {
            draw.draw_card(p1).unwrap();
            draw.draw_card(p2).unwrap();
            draw.draw_card(p3).unwrap();
            draw.draw_card(p4).unwrap();
            draw.draw_card(p5).unwrap();
            draw.draw_card(p6).unwrap();
            draw.draw_card(p7).unwrap();
            draw.draw_card(p8).unwrap();
        }

        // p1 bids and wins, trump is now Spades and 7s.
        assert!(draw.bid(p1, cards::S_7, 1));

        let mut exchange = draw.advance(p1).unwrap();
        let friends = vec![
            FriendSelection {
                card: cards::D_3,
                initial_skip: 0,
            },
            FriendSelection {
                card: cards::D_4,
                initial_skip: 0,
            },
            FriendSelection {
                card: cards::D_5,
                initial_skip: 0,
            },
        ];
        exchange.set_friends(p1, friends).unwrap();
        let mut play = exchange.advance(p1).unwrap();
        match play.game_mode {
            GameMode::FindingFriends { num_friends: 3, .. } => (),
            _ => panic!("Didn't have 3 friends once game was started"),
        }

        assert_eq!(
            play.landlords_team,
            vec![p1],
            "Nobody should have joined the team yet"
        );

        // Play the first hand. P2 will join the team.
        play.play_cards(p1, &p1_hand[..1]).unwrap();
        play.play_cards(p2, &p2_hand[..1]).unwrap();
        play.play_cards(p3, &p3_hand[..1]).unwrap();
        play.play_cards(p4, &p4_hand[..1]).unwrap();
        play.play_cards(p5, &p5_hand[..1]).unwrap();
        play.play_cards(p6, &p6_hand[..1]).unwrap();
        play.play_cards(p7, &p7_hand[..1]).unwrap();
        play.play_cards(p8, &p8_hand[..1]).unwrap();

        // Check that P2 actually joined the team.
        let msgs = play.finish_trick().unwrap();
        assert_eq!(
            msgs.into_iter()
                .filter(|m| match m {
                    MessageVariant::JoinedTeam { player } if *player == p2 => true,
                    _ => false,
                })
                .count(),
            1
        );

        assert_eq!(play.landlords_team, vec![p1, p2]);

        // Play the next trick, where the landlord will join the team, and then
        // p2 will join the team (again).
        play.play_cards(p1, &p1_hand[1..2]).unwrap();
        play.play_cards(p2, &p2_hand[1..2]).unwrap();
        play.play_cards(p3, &p3_hand[1..2]).unwrap();
        play.play_cards(p4, &p4_hand[1..2]).unwrap();
        play.play_cards(p5, &p5_hand[1..2]).unwrap();
        play.play_cards(p6, &p6_hand[1..2]).unwrap();
        play.play_cards(p7, &p7_hand[1..2]).unwrap();
        play.play_cards(p8, &p8_hand[1..2]).unwrap();

        // We don't get any joined-team messages, because both team-joiners have
        // already joined.
        let msgs = play.finish_trick().unwrap();
        assert_eq!(
            msgs.into_iter()
                .filter(|m| match m {
                    MessageVariant::JoinedTeam { .. } => true,
                    _ => false,
                })
                .count(),
            0
        );

        // Assert that the team didn't get any bigger
        assert_eq!(play.landlords_team, vec![p1, p2]);
        // But also that all of the friend cards have been played!
        match play.game_mode {
            GameMode::FindingFriends { ref friends, .. } => assert!(
                friends.iter().all(|f| f.player_id.is_some()),
                "all friends lots taken"
            ),
            _ => unreachable!(),
        }

        // Finish the game; we should see the landlord go up 4 levels (3 for
        // keeping the opposing team at 0, and a bonus level)

        let (new_init_phase, msgs) = play.finish_game().unwrap();
        assert_eq!(
            msgs.into_iter()
                .filter(|m| match m {
                    MessageVariant::BonusLevelEarned => true,
                    MessageVariant::RankAdvanced { player, new_rank } if *player == p1 => {
                        assert_eq!(*new_rank, Number::Jack);
                        false
                    }
                    MessageVariant::RankAdvanced { player, new_rank } if *player == p2 => {
                        assert_eq!(*new_rank, Number::Six);
                        false
                    }
                    _ => false,
                })
                .count(),
            1
        );

        assert_eq!(
            new_init_phase
                .propagated
                .players
                .into_iter()
                .map(|p| p.level)
                .collect::<Vec<Number>>(),
            vec![
                Number::Jack,
                Number::Six,
                Number::Two,
                Number::Two,
                Number::Two,
                Number::Two,
                Number::Two,
                Number::Two
            ],
            "Check that propagated players have the right new levels"
        );
    }

```

What do you think? (it also asserts more about the final state of the game).

core/src/interactive.rs
```diff
@@ -386,7 +393,8 @@ impl BroadcastMessage {
             ThrowEvaluationPolicySet { policy: ThrowEvaluationPolicy::All } => format!("{} set throws to be evaluated based on all of the cards", n?),
             ThrowEvaluationPolicySet { policy: ThrowEvaluationPolicy::Highest } => format!("{} set throws to be evaluated based on the highest card", n?),
             RevealedCardFromKitty => format!("{} revealed a card from the bottom of the deck", n?),
-            GameFinished { result: _ } => "The game has finished.".to_string()
+            GameFinished { result: _ } => "The game has finished.".to_string(),
+            BonusLevelEarned => "Landlord team earned a bonus level for winning with a smaller team".to_string(),
```
rbtying, line 397: nit: "Defending team"?

core/src/game_state.rs
cai-jack: Yes. It does. 

I created the test case by actually playing through the entire game. My intention was to use this as a template for other scenarios that may need to be tested, as close to actual play as possible, I just rename it as test_full_game_play.

PR #166, opened 2020-06-29
Add a game separator message to visually separate a new game from pre…

cai-jack:
…vious one in the chat box.

people tend to click on finish game and start right away, making it difficult to read about last game's summary.

We may need to improve this so that just finished game can be better summarized and presented. But for now, this will make it easier so that no one needs to type in something  like "=========" quickly before a new game starts

### DIFF
--- a/core/src/interactive.rs
+++ b/core/src/interactive.rs
@@ -75,10 +75,7 @@ impl InteractiveGame {
             (Message::StartGame, GameState::Initialize(ref mut state)) => {
                 info!(logger, "Starting game");
                 self.state = GameState::Draw(state.start()?);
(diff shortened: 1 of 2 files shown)


rbtying:
Rather than adding a game-separator message to be emitted from the backend, can we update the frontend to handle the render Starting Game message better, similar to how we do for played cards?

You can add a case here:
https://github.com/rbtying/shengji/blob/master/frontend/src/ChatMessage.tsx#L26

though you might need to update the type listing in types.ts to include StartingGame

cai-jack:
use front end chat msg rendering instead

Issue #68, opened 2020-04-13
Option to disable take-backs for game


ezhan94:
Prevents players from exploiting it to test throws. Since you can queue up a play anyways, this feature isn't necessary imo.

rbtying:
All other players are notified via game-chat that a take back has occurred, so I don't think the ability to take back a play makes a meaningful difference here?

The implementation currently assumes that players want to play in good faith. For example, there's nothing stopping you from opening a new tab and entering the room using somebody else's name to see their hand.

cai-jack:
just pushed in the codes for this. This actually made a difference in our play. U can't always assume good faith playing in all cases, as it turned out :)

Also, I discovered a nice feature when takeback is disabled. Everytime when a trick is blocked, we took back the play, replayed the blocked cards. As it turned out, you don't actually have to do that. With takeback disabled, I found out following players can only play against the blocked cards! Save a bit of hustle in playing, the previous trick shows the blocked trick, and makes implementing this feature much simpler (I was thinking that I needed to handle this case after takeback is disabled)

PR #169, opened 2020-07-03
#68: add game option to disable play take back

cai-jack:


core/src/interactive.rs
```diff
@@ -392,6 +397,8 @@ impl BroadcastMessage {
                 protected longer tuples from being drawn out by shorter ones (pair does not draw triple)", n?),
             ThrowEvaluationPolicySet { policy: ThrowEvaluationPolicy::All } => format!("{} set throws to be evaluated based on all of the cards", n?),
             ThrowEvaluationPolicySet { policy: ThrowEvaluationPolicy::Highest } => format!("{} set throws to be evaluated based on the highest card", n?),
+            PlayTakebackPolicySet { policy: PlayTakebackPolicy::AllowPlayTakeback } => format!("{} allowed play takeback", n?),
```
rbtying, line 400: "taking back plays" rather than "play takeback"?

PR #178, opened 2020-07-03
bug fix: eliminate flicker during drawing by using synced file loadin…

cai-jack:
…g instead of async

the use of suspense and lazy loading of svg cards causes annoying flickering especially at draw phase. This fix should make it  visually more pleasant to use

### DIFF
--- a/frontend/src/Card.tsx
+++ b/frontend/src/Card.tsx
@@ -5,327 +5,114 @@ import InlineCard from "./InlineCard";
 import { cardLookup } from "./util/cardHelpers";
 import { SettingsContext } from "./AppStateProvider";
 
-const Svg1B = React.lazy(
-  async () => await import("./generated/playing-cards/1B")
-);
-const Svg2C = React.lazy(
-  async () => await import("./generated/playing-cards/2C")
-);
-const Svg2J = React.lazy(
-  async () => await import("./generated/playing-cards/2J")
-);
-const Svg3D = React.lazy(
-  async () => await import("./generated/playing-cards/3D")
-);
-const Svg4C = React.lazy(
-  async () => await import("./generated/playing-cards/4C")
-);
-const Svg4S = React.lazy(
-  async () => await import("./generated/playing-cards/4S")
-);
-const Svg5H = React.lazy(
-  async () => await import("./generated/playing-cards/5H")
-);
-const Svg6D = React.lazy(
-  async () => await import("./generated/playing-cards/6D")
-);
-const Svg7C = React.lazy(
-  async () => await import("./generated/playing-cards/7C")
-);
-const Svg7S = React.lazy(
-  async () => await import("./generated/playing-cards/7S")
-);
-const Svg8H = React.lazy(
-  async () => await import("./generated/playing-cards/8H")
-);
-const Svg9D = React.lazy(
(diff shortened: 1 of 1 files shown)


cai-jack:
hmm, ci failed complaining about something I did not change. local cargo clippy runs  fine:

    Checking shengji-core v0.1.0 (/Users/rust/Projects/shengji/core)
    Checking shengji v0.1.0 (/Users/rust/Projects/shengji/backend)
    Finished dev [unoptimized + debuginfo] target(s) in 8.97s

cai-jack:
error: couldn't read backend/src/../../frontend/dist/playing-cards.js: No such file or directory (os error 2)
##[error]   --> backend/src/main.rs:688:25
    |
688 | static CARDS_JS: &str = include_str!("../../frontend/dist/playing-cards.js");
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

looks like ci does not pull this file?

cai-jack:
@rbtying

cai-jack:
yes. it is consistently flickering during the draw phase, even after resetting the game with cards already loaded, not very usable in my opinion. Also, the svg files are already loaded into client side with the static imports, should not consume extra bandwidth afterwards.

rbtying:
> yes. it is consistently flickering during the draw phase, even after resetting the game with cards already loaded, not very usable in my opinion.
Interesting. we could try bundling all of the SVGs together into a single async loaded component?

> Also, the svg files are already loaded into client side with the static imports, should not consume extra bandwidth afterwards.
Since 90% of players will never use the SVG cards, and the JS bundle without the SVG is ~350KiB (and with SVG is ~1.4MiB), I'd prefer not to slow down load time for all players unless they have actually opted into using SVG cards. I can't reproduce the issue you're seeing -- what browser are you using? Do you see more than one request to playing-cards.js?

cai-jack:
Google Chrome on MacOS. I ran a few more tests: If a subsequent game draws cards that were not in previous game(s), the most noticeable flickering occurs when a "new" card is drawn and inserted into the middle of the hand. It feels like the cards after the insertion point jumped for the insertion:). I don't think it is the issue of more request to playing-cards.js, the async loading is. If your concern is the bandwidth, is there a way to conditionally load them and make the cards available only when the setting is on? react.lazy seems to make sure you can't do that until the component is ready to render

rbtying:
Interesting. Like I said, I can't reproduce -- I tried with Chrome on Catalina. The intent of the `React.lazy` here was to async load all of the cards once any card was requested. If we're for some reason observing transient states while loading new cards, even after the network request, we can try bundling all of the cards into a single component to async load (rather than loading them one at a time, even though the underlying file is the same file). If you're familiar with another way to do this, I'm happy to try it?

I'm strongly against adding ~3 seconds of load time (@ 300KiB/s download rate) for a feature that very few people seem to use.

rbtying:
https://github.com/rbtying/shengji/tree/svg-card-batch-load

Does this resolve the issue locally for you ?

cai-jack:
Yes. it does! Thanks.

PR #180, opened 2020-07-08
remove a user's prior connection if the user joins the same game again

cai-jack:
to prevent spoofing during game play. All users will also be alerted about the repeated connection


rbtying:
Hm. This had been an intentional feature to allow shadowing for new players (and to handle random internet disconnects)

Has this been an issue in your game play? we could make it configurable in the settings panel

cai-jack:
Ok. let's make it an option. It has not because no one knows about it yet, but I have the sinking feeling that it will make the game unplayable once it becomes known :)

cai-jack:
btw, do u know of a better IDE for rust than vscode? I can't drill down to macros with vscode. Thanks.

rbtying:
I think this is one of those things where it would be problematic if you play with strangers, but the game is pretty much always played with friends, so it is a non-issue.

re: IDE. I use neovim, mostly. But this codebase is fairly macro-light? (We can probably remove the one macro in game_state.rs and replace it with `x.ok_or_else(|| anyhow!("unwrapped none?"))?`). Have you tried using the rust-analyzer VS Code plugin rather than the rls-based one? It might work better for you

cai-jack:
I don't have the same confidence regarding game playing -  the main motivation for me to add the no takebacks on bidding and play :)

Thanks. I will give the rust-analyzer a try.

In vscode, I can't drill down to info! for example, looks like a logging related macro, but I have no way of knowing where it is even coming from

rbtying:
ah -- info! /etc are imported at the top of the file from `slog`. It's surprising to me that VS Code won't link you there, though?

cai-jack:
cool, rust-analyzer does! I am still getting used to the extra type info injected into the code :)

backend/src/main.rs
```diff
@@ -512,6 +512,19 @@ async fn user_connected(ws: WebSocket, games: Games, stats: Arc<Mutex<InMemorySt
             };
             info!(game.tracer(&logger, &room, Some(1)), "Joining room"; "player_id" => player_id.0);
             game.users.insert(ws_id, UserState { player_id, tx });
+
+            // if the same user joined before, remove its previous entry from the user list
+            if game
+                .game
+                .dump_state()
```
rbtying, line 519: Can we add an accessor function for this field directly rather than cloning the entire game structure?

backend/src/main.rs
```diff
@@ -512,6 +512,19 @@ async fn user_connected(ws: WebSocket, games: Games, stats: Arc<Mutex<InMemorySt
...
+                .unwrap()
```
rbtying, line 520: unwraps cause the server to panic -- let's avoid that and pick a safe default?

frontend/src/Initialize.tsx
```diff
@@ -878,6 +898,22 @@ const Initialize = (props: IProps): JSX.Element => {
...
         </div>
+        <div>
+          <label>
+            User joining game policy:{" "}
+            <select
+              value={props.state.propagated.user_multi_game_session_policy}
+              onChange={setUserMultiGameSessionPolicy}
+            >
+              <option value="AllowMultipleSessions">
+                A user can join the same game with multiple sessions
```
rbtying, line 909: maybe, "Users can join the game in multiple tabs"

frontend/src/Initialize.tsx
```diff
@@ -878,6 +898,22 @@ const Initialize = (props: IProps): JSX.Element => {
...
+              </option>
+              <option value="SingleSessionOnly">
+                A user can join the same game with only one session
```
rbtying, line 912: and, "Users can only join the game in one tab"

backend/src/main.rs
```diff
@@ -616,8 +629,30 @@ async fn user_connected(ws: WebSocket, games: Games, stats: Arc<Mutex<InMemorySt
...
+                    ) = m
+                    {
+                        user_single_game_session = true;
+                    } else {
+                        user_single_game_session = true;
+                    }
+
                     match game.game.interact(m, player_id, &logger) {
                         Ok(msgs) => {
+                            if user_single_game_session {
```
rbtying, line 644: hm, why do we need to check this on every interaction? Dropping existing users when a new one joins seems like it should be sufficient, right?

cai-jack: this check is only needed when the action is SetUserMultiGameSessionPolicy with SingleSessionOnly. We have to scan the users to reduce existing users with multiple sessions to just one. The ideal place will be inside either interactive or game_state, but then we have to pass down users, which is only defined in main.rs. Please let me know how you'd like to handle. (We can also just not do anything at all to make the code look cleaner if u'd like)

cai-jack: also, any reason why users is not part of the game state?

rbtying: Websocket connections aren't part of the game state because the `shengji-core` library is intended to be agnostic to how the players actually interact with the backend. The `backend` code does the appropriate mapping in both directions to make it work.

I'd prefer not to do this check here, and instead update the phrasing on the setting in the UI to say something like

'allow players to be shadowed by joining with the same name"

backend/src/main.rs
```diff
@@ -512,6 +512,15 @@ async fn user_connected(ws: WebSocket, games: Games, stats: Arc<Mutex<InMemorySt
             };
             info!(game.tracer(&logger, &room, Some(1)), "Joining room"; "player_id" => player_id.0);
             game.users.insert(ws_id, UserState { player_id, tx });
+
+            // if the same user joined before, remove its previous entry from the user list
+            if game.game.get_user_multi_game_session_policy()
```
rbtying, line 517: nit: let's do

```
if game.game.allows_multiple_sessions_per_user() {
}
```
and avoid looking at the `UserMultiGameSessionPolicy` enum at all

core/src/game_state.rs
```diff
@@ -195,6 +195,18 @@ impl Default for KittyTheftPolicy {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum UserMultiGameSessionPolicy {
```
rbtying, line 199: Let's call this `ShadowingPolicy` or sth

backend/src/main.rs
cai-jack: check is removed as well

PR #181, opened 2020-07-10
restrict game start to the landlord unless no landlord is determined

cai-jack:
the request is landlord should have the exclusive privilege to start the game during play. (currently, the end game notification indicates the landlord will start the game, but anyone can). Some newly minted landlord feels HER hard earned victory was not sufficiently rewarded without the exclusive privilege to start a new game:)

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -2090,7 +2090,7 @@ mod tests {
         let p2 = init.add_player("p2".into()).unwrap().0;
         let p3 = init.add_player("p3".into()).unwrap().0;
         let p4 = init.add_player("p4".into()).unwrap().0;
-        let mut draw = init.start().unwrap();
+        let mut draw = init.start(PlayerID(0)).unwrap();
         // Hackily ensure that everyone can bid.
         draw.deck = vec![
             cards::S_2,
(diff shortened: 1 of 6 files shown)


rbtying:
Oh, interesting. this is just restricting the "start game" button?

The "starting the game" from the message was intended to be "first to draw" and "first to play", haha

cai-jack:
yes:)

cai-jack:
the drawing does start with the landlord in the UI, otherwise it starts with the first player

core/src/game_state.rs
```diff
@@ -1778,11 +1778,23 @@ impl InitializePhase {
         }
     }
 
-    pub fn start(&self) -> Result<DrawPhase, Error> {
+    pub fn start(&self, id: PlayerID) -> Result<DrawPhase, Error> {
         if self.propagated.players.len() < 4 {
             bail!("not enough players")
         }
 
+        match self.propagated.landlord {
```
rbtying, line 1786: maybe

```
if self.propagated.landlord.map(|l| l != id).unwrap_or(false) {
   bail!("Only the landlord can start the game")
}
```

cai-jack: cool!

cai-jack:
understood. It is a desire by our player to have the game match the actual physical play a bit closer :)

cai-jack:
Looks like this feature is actually in demand in our circle. I turned it into an option. Hope this is acceptable to you. Thanks.

Issue #21, opened 2020-04-05
Implement endgame "you win" after defending on A


rbtying:
Right now, nothing happens when you win, and the level is capped at A.

It should probably be capped at NT, and there should be some indication of winning

PR #184, opened 2020-07-13
#21: add ticker tape and fireworks to announce game set winner

cai-jack:
see screenshot, after a landlord successfully defend A.

Screen Shot 2020-07-12 at 5 47 54 PM

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -220,6 +220,8 @@ pub struct PropagatedState {
     pub players: Vec<Player>,
     pub observers: Vec<Player>,
     landlord: Option<PlayerID>,
+    gameset_winner_declared: bool,
+    gameset_winner_player_id: Option<PlayerID>,
     #[serde(default)]
     landlord_emoji: Option<String>,
     chat_link: Option<String>,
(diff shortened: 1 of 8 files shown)


core/src/game_state.rs
```diff
@@ -220,6 +220,8 @@ pub struct PropagatedState {
     pub players: Vec<Player>,
     pub observers: Vec<Player>,
     landlord: Option<PlayerID>,
+    gameset_winner_declared: bool,
```
rbtying, line 223: Can this be a message rather than a game state?

(if it has to be a game state, we should model it as a more complete enum, rather than as two fields)

cai-jack: We can. Logically it makes more sense as well since it should be notification, not state. But that will turn ChatMessage into a notification handler, kind of messy.

Let me use a separate pull request to do it. 

we can decide afterwards.

rbtying: You can add another websocket handler if that's helpful -- this is what we do for game statistics, for example.

cai-jack:
The ticker tape goes away after a player clicks the start game button, game plays on just like before.

confetti only makes using message much simpler I think. Will get that working first.

cai-jack:
use message instead of game state for game set winner announcement

PR #185, opened 2020-07-13
Game set winner using message

cai-jack:


cai-jack:
@rbtying no sure why ci was not triggered

core/src/game_state.rs
```diff
@@ -1158,7 +1158,7 @@ impl PlayPhase {
         msgs
     }
 
-    pub fn finish_game(&self) -> Result<(InitializePhase, Vec<MessageVariant>), Error> {
+    pub fn finish_game(&mut self) -> Result<(InitializePhase, Vec<MessageVariant>), Error> {
```
rbtying, line 1161: Hm. This shouldn’t mutate existing state? We are just sending a message, right?

core/src/game_state.rs
```diff
@@ -1209,6 +1209,18 @@ impl PlayPhase {
             msgs.push(MessageVariant::BonusLevelEarned);
         };
 
+        let current_landlord_idx = bail_unwrap!(self
+            .propagated
+            .players
+            .iter()
+            .position(|p| p.id == self.landlord));
+
+        if landlord_won && self.propagated.players[current_landlord_idx].rank() == Number::Ace {
```
rbtying, line 1218: I think we already create a message at the end of each game? Can we add a field to that, or see if it is already included?

core/src/game_state.rs
cai-jack: good catch, used for setting game state in the other branch

core/src/game_state.rs
cai-jack: GameFinished message is for every game's result. It is only used to generate game finished message. All other messages on player advances are separately generated. I think it is much cleaner to have a separate message for Gameset winner, it makes ChatMessage handling much simpler

cai-jack:
still no ci run, r u updating ci vm? @rbtying

rbtying:
the CI is via GitHub Actions— it looks like its queueing quite a bit right now. Might be a transient issue; I haven’t made changes for a while

core/src/game_state.rs
```diff
@@ -1788,7 +1801,7 @@ impl InitializePhase {
         }
     }
 
-    pub fn start(&self) -> Result<DrawPhase, Error> {
+    pub fn start(&mut self) -> Result<DrawPhase, Error> {
```
rbtying, line 1804: nit: no `mut` here, either

core/src/interactive.rs
```diff
@@ -457,6 +457,7 @@ impl BroadcastMessage {
             PutDownCards => format!("{} put down the bottom cards", n?),
             GameFinished { result: _ } => "The game has finished.".to_string(),
             BonusLevelEarned => "Landlord team earned a bonus level for defending with a smaller team".to_string(),
+            GamesetWinnerAnnoucement {ref player_name} => format!("{} won the game set!", player_name),
```
rbtying, line 460: "{} successfully defended on A!", instead? "game set" isn't (as far as I know) a common term in the game.

frontend/src/ChatMessage.tsx
```diff
@@ -41,7 +46,16 @@ interface IProps {
   message: IMessage;
 }
 const ChatMessage = (props: IProps): JSX.Element => {
+  const [fireworkStarted, setFireworkStarted] = React.useState<boolean>(false);
```
rbtying, line 49: Can we factor the `firework` into its own component, similar to the Beeper? Then, we can lazily include the firework on the associated with the winner announcement, and the firework can manage its own state.

frontend/src/firework.ts
```diff
@@ -0,0 +1,49 @@
...
+    confetti(
+      Object.assign({}, defaults, {
+        particleCount,
+        origin: { x: randomInRange(0.7, 0.9), y: Math.random() - 0.2 },
+      })
+    );
+  }, 250);
+
+  setTimeout(() => {
+    handler = window.addEventListener("click", (event) => {
```
rbtying, line 45: This will leak event listeners on the Window, I think? Since there's no guarantee that we'll re-run this function after `clearInterval` is called.

I'd ~strongly prefer that this is rewritten as a React component

core/src/interactive.rs
cai-jack: I just google up other card games. Seems like the most common way to describe trick taking games is to name the entire series as one game, and each "play" as one deal. so winners win the game by advancing their ranks/levels in each deal until A is successfully defended. Since game is used in place of deal, it is bit hard to come up with a common term. 

"successfully defended on A" does not carry the significance of winning the game, I think. 

"{} won the game series"? or simply "Congratulations! {} is the winner!"?

frontend/src/ChatMessage.tsx
cai-jack: Yes. we can do that. The situation for Beeper is quite different from firework though. firework is a response to an event, whereas Beeper is for config setting. To properly handle server generated events, we will eventually need to have some kind of event handling capacities on both client and server side. ChatMessage is for rendering the messages. It actually causes firework to fire each time the chat box renders its messages without hacking the rendering. Since the underlying mechanism is essentially a state, hacking it to handle events will not look that great in the long run.

frontend/src/firework.ts
cai-jack: You are right. Will fix. The listener is meant for players to stop the firework before it runs its cause.

core/src/game_state.rs
cai-jack: same deal, will fix

frontend/src/ChatMessage.tsx
rbtying: There are event handling capabilities in the websocket handler -- it probably wouldn't be difficult for that to create a react event for a hook. But, modeling the firework as a component that renders fireworks while it is rendered should make this moot? IIUC idiomatic React prefers to minimize "events" in favor of state-rendering

cai-jack: There is a difference in moving the firework to the component three. The current hack ensures each winning message generates one and only one firework by associating firework with each winning message (an instance of ChatMessage). When another winning message arrives, it will trigger another firework (from another instance of ChatMessage). When we create a firework component in the component tree, simply duplicating the logic will have the effect of generating one firework regardless of number of winning messages received. If we use state to handle when to trigger firework, then we are back to where we were before this in previous branch. The firework component will need to be able to somehow generate the firework exactly once for each winning message. 

question to you, if I use the key attribute with different value, will react create a different component instance even though the component is not part of a list? Thanks.

rbtying: > question to you, if I use the key attribute with different value, will react create a different component instance even though the component is not part of a list? Thanks.

Honestly, not sure. But hopefully not relevant?

> the current hack ...

What I am thinking is to have the component set up the firework animation (including setInterval, etc) in `useEffect`, and then use the function-return-value from `useEffect` to implement tearDown.

Something like this:

```
const Firework = (props) => {
  const { setInterval, clearInterval } = React.useContext(TimerContext);
  const { elapsed, setElapsed } = React.useState(0);
  React.useEffect(() => {
     // set up handlers
    const t = setInterval(() => {
      // ...
      setElapsed(Date.now());
    }, 100);
    return () => {
      clearInterval(t);
      // clean up other handlers
    }
  }, [])

  return (<>
    <canvas...>
  </>);
}
```

cai-jack: See my changes in the upcoming commit. Don't think the same approach as Beeper works since it does not resolve the issue of when to trigger the fireworks.

The Firework component is just a simple wrap to make ChatMessage cleaner, and still adhere to the semantics of event handling.

Also, no clean up is needed for confetti as it cleans up by itself, and the instance can be reused when startFireworks is called again (if ever)

rbtying: https://www.npmjs.com/package/react-confetti-canvas

It looks like other folks have already implemented this :)

cai-jack: I did see that. Yeah, don't think you will like it. 

It is 3.7M unpacked vs 61.1k.

https://www.npmjs.com/package/canvas-confetti

and weekly downloads are 100+ times more for the non-react version :)

rbtying: I think the vast majority of the size there is the gif in the package preview. The actual source is pretty short.

https://github.com/fufu70/react-confetti-canvas/blob/master/assets/canvas.gif <-- this thing

Anyways, I can poke at this in the evening at the end of the work day.

cai-jack: sounds good. The main feature I like about the non-react version is the ability to put the confetti all over the page instead of only in a separate panel. The firework code actually comes from the author's very cool demo page.

https://www.kirilv.com/canvas-confetti/

check it out if you can

cai-jack:
use rying's impl instead

PR #186, opened 2020-07-14
Add confetti/full screen notification on win

### LINKED ISSUE #21: Implement endgame "you win" after defending on A

rbtying:
Fixes #21; based on #185

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -896,6 +896,7 @@ pub struct PlayerGameFinishedResult {
     pub is_defending: bool,
     pub is_landlord: bool,
     pub ranks_up: usize,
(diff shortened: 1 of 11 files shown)


rbtying:
@cai-jack what do you think about this approach? I think it makes the state management considerably simpler...

cai-jack:
Looks pretty neat! back to state based:) Good for me.

Change the font color of h1 in confetti to red?

rbtying:
It's a little hard to read in red (b/c some confettis are red). But, there's no modification to game state here -- just app state.

Issue #190, opened 2020-07-18
Semantics of successfully defending A in "Finding Friends" game?


cai-jack:
for a player to be said to defend A successfully, our understanding is the player should be the landlord, not just on the landlord team. This is mainly a problem with the "Finding Friends" style of playing, since landlord team players can have different ranks.

rbtying:
Hm. I think this is up to interpretation, but my mind has been; if you are on the team, you started on A, and you go up a level, it counts -- this is consistent with fixed-team semantics, where either player successfully landlord-ing on A counts.

WDYT?

cai-jack:
That is one way of seeing it. It just feels weird to us when the landlord won on say, 7, and the other team player got a congrats on defending A :)

Any objection of me adding an option to only allow landlord to win on defending A, for the finding friend style only?

rbtying:
ah, I see. fair enough. We can change the default here, so that (at least) the landlord has to defend A

PR #198, opened 2020-07-20
#190: in finding friends game, only landlord can defend A

cai-jack:


core/src/game_state.rs
```diff
@@ -1139,7 +1140,12 @@ impl PlayPhase {
                 for bump_idx in 0..bump {
                     match advancement_policy {
                         // Player *must* defend on Ace and win to advance.
-                        _ if player.rank() == Number::Ace && !(is_defending && bump_idx == 0) => {
+                        _ if player.rank() == Number::Ace
+                            && !(is_defending
+                                && (!finding_friends_game
```
rbtying, line 1145: I think if all of the defending team is rank A, then they should all win, no?

probably the right way to do this is to explicitly check if the landlord successfully defended A in the first loop, and then allow advances past A if the player started at A and is on the landlord's team.

This also removes the need for the check on `finding_friends_game`, since the rule would work for both tractor and finding friends

cai-jack:
hmm, my cargo clippy has not complaints. The errors in ci clippy does not make sense. @rbtying

rbtying:
did you run `cargo clean -p $crate` for the crate you want to check clippy for? `cargo clippy` will noop if the binary is already built

cai-jack:
no complaints after I ran cargo clean -p shengji, then cargo clippy on my side.

core/src/game_state.rs
cai-jack: That will block landlord with A even if the landlord wins if there is a player with non-A rank.

We can make players on landlord team with rank A to pass when landlord with A wins

cai-jack:
oh, the lint error came from shengji-core. But why it is limiting the number of parameters???

core/src/game_state.rs
rbtying: ? the criteria would be

```
const landlord_successfully_defended_a = ...;

if player.rank() == Number::Ace && !(is_defending && bump_idx == 0 && landlord_successfully_defended_a) =>

```

rbtying:
> oh, the lint error came from shengji-core. But why it is limiting the number of parameters???

The linter prefers to have fewer parameters per function (you can move the parameters to a struct) to help with readability, in general. But, I don't think you need the extra parameter anyways?

cai-jack:
I need to pass in landlord_level instead. finish_game already has the calculation, don't want to repeat again in the function. I disabled the check for this function.

cai-jack:
still getting used to how tests are compiled. The compiler skips compiling test codes in build even if they are in the same file as the regular codes!

rbtying:
> still getting used to how tests are compiled. The compiler skips compiling test codes in build even if they are in the same file as the regular codes!

It will compile them if you do `cargo build --tests` or `cargo check --tests`, as well as (obviously) in `cargo test`

cai-jack:
It is not that, (I think) it is the calculation of landlord_idx. It is already available, don't want to repeat

rbtying:
Can you pass `landlord: (PlayerID, Number)` then?

cai-jack:
> > still getting used to how tests are compiled. The compiler skips compiling test codes in build even if they are in the same file as the regular codes!
> 
> It will compile them if you do `cargo build --tests` or `cargo check --tests`, as well as (obviously) in `cargo test`

default to build and run tests will be better

core/src/game_state.rs
cai-jack: I could not figure out why you checked !is_defending to block advance when checking should be on is_defending. The condition I use is more intuitive I think.

rbtying:
> default to build and run tests will be better

just use `cargo test` then? or do you mean for `cargo clippy`?

core/src/game_state.rs
cai-jack: I did test playing to make sure the updated check condition works.

core/src/game_state.rs
```diff
@@ -1139,7 +1139,10 @@ impl PlayPhase {
                 for bump_idx in 0..bump {
                     match advancement_policy {
                         // Player *must* defend on Ace and win to advance.
-                        _ if player.rank() == Number::Ace && !(is_defending && bump_idx == 0) => {
+                        _ if landlord.1 != Number::Ace
+                            && player.rank() == Number::Ace
+                            && is_defending =>
```
rbtying, line 1144: the check as written is incorrect:

1. attacking team can advance past A (!is_defending)
2. if landlord is defending ace, defending team at K can skip past A

rbtying: rule is:

1. all players who did not start on Ace cannot advance past Ace
2. if the player is on the defending team, and if the landlord successfully defended Ace, then the player can advance past Ace

I don't mind if you rewrite the condition here, but the truth table has to be the same

cai-jack: > the check as written is incorrect:
> 
> 1. attacking team can advance past A (!is_defending)

> 2. if landlord is defending ace, defending team at K can skip past A

you are right, my bad. Will revert to your check

cai-jack: did not realize player's rank was updated inside the loop...

cai-jack:
> > default to build and run tests will be better
> 
> just use `cargo test` then? or do you mean for `cargo clippy`?

I meant it will be better for cargo build to build and run tests as default without extra parameter, similar to maven's build in which test by default is an included phase

rbtying:
Thanks for fixing this!

Issue #197, opened 2020-07-20
Use WASM code in frontend to generate state variables


rbtying:
- [x] `cards` can be generated from `hands`
- [x] Players who are not starting the trick can be told immediately whether their play is legal or not
- [x] Scoring analysis should be shared between frontend and backend
- [ ] Probably want a top-level state item for the current playerID -- linearly scanning for it in every place that needs it feels silly

PR #207, opened 2020-07-26
Use WASM to sort and group cards from the hands structure

rbtying:
Rather than encoding cards twice (in a fairly expensive way!) on the
wire, use code in the frontend to generate the appropriate list-of-cards
format. Previously, we didn't do this because the sort order of the
cards was undefined without additional state; now, we can do better by
using logic in WASM to sort.


rbtying:
part of #197

PR #223, opened 2020-08-14
bug fix: landlord is a playerId, not an index to the player array

cai-jack:


cai-jack:
it works until a player is moved in the player list...

cai-jack:
nah, it is me. but thanks for taking the credit though :)

PR #230, opened 2020-09-05
Fix bug where kitty is not sent to chat if it contains no points

### LINKED ISSUE #228: Print out the cards from the bottom pile after the game has ended in chat
Found that sometimes we miss seeing what cards were in the bottom pile when revealed after the last trick is played because someone will often click the "end game" button right away. It would be nice if the cards were printed in the chat so we could still see what they were even after the game has ended.

alvinsiu:
Fixes #228

Note: I did not attempt to run this locally, so this code is untested.


rbtying (approved):
thanks for the fix!

Issue #222, opened 2020-08-13
Allow more step sizes and offsets for custom scoring


chnlyi:
Liked the recent changes for  #221! The colored bar is my favorite!

However, would it be more intuitive to define the step/steps as points plus or minus the defeating point?  also, can more steps (20 30 40 50 60 70 80 90 100) be added?

For example, if there are 5 guys playing 2 decks, first, they define the defeating point as 90 (not 80 because they feel landlord only has one friend, or whatever...); then they want to increase the speed of shengji so that they define the step to be 30 instead of 40. Now the tie (no side advance) will be 90 to 120; landlord advances 1 level when the score is between 60 and 85; attacking advances 1 level when the score is between 120 and 150; landlord advances 2 levels when the score is between 30 and 55; attacking advances 2 levels when the score is between 150 and 180; etc

rbtying:
Worth considering! The main issue I ran into was that the default scoring is not actually just linear in steps: it's not 0-35, 40-75, 80-115, 120-155, 160-195, 200; it's actually 0, 5-35, ...

As a result, people are likely going to be surprised if either 0 is treated specially (potentially truncating the first step by more than 5 points, if the offset is allowed) or if it isn't (doesn't map to the default game well).

So, the initial pass was to only allow step sizes which evenly divide 100, and to force steps to start at zero. Curious to hear your thoughts on what the "right" zero-crossing / max-points-crossing behavior would be

xiaofan-iit:
Would support more flexibility of step sizes around the default value. For example, for 3 decks of cards, the default size is 60; could you add 50, 55, 65, 70 as choices of step size? Similarly, for other numbers of decks. The reason for the suggestion is that the number of players, the number of friends and the number of bottom cards change. Thus, we would like to adjust the step size around the default value (up and down upto 15 points). Thanks.

rbtying:
@xiaofan-iit that makes sense -- the current limitation was predominantly to make it so that the scoring scales with the number of decks (so, the step-size is the number of decks multiplied by a deck-invariant step). It also currently requires that the deck-invariant step evenly divides a single deck.

Do you mind filing another issue for that? I think it's a different issue

xiaofan-iit:
@rbtying Thanks for your consideration. I will file as a different issue (if I can find how to file a new issue:)).

PR #246, opened 2020-11-02
Support more granular step sizes in scoring settings

### LINKED ISSUE #245: Allowing more flexible step sizes
Under 'Scoring settings', could you make 'Step size' to be more flexible? We would like more flexibility of step sizes around the default value. For example, for 3 decks of cards, the default size is 60; could you add 50, 55, 65, 70 as choices of step size? Similarly, for other numbers of decks. The reason for the suggestion is that the number of players, the number of friends and the number of bottom cards change. Thus, we would like to adjust the step size around the default value (up and down upto 15 points). Thanks.

rbtying:
Fixes #245

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -583,8 +583,9 @@ impl PlayPhase {
             landlord_won,
             landlord_bonus: bonus_level_earned,
         } = compute_level_deltas(
-            propagated.game_scoring_parameters,
+            &propagated.game_scoring_parameters,
             self.num_decks,
+            crate::scoring::POINTS_PER_DECK,
             non_landlords_points,
             smaller_landlord_team,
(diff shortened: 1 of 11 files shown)


rbtying:
Also helps with #222, though not completely better for that

xiaofan-iit:
Thanks for making the modification. However, the 'Scoring settings' does not work now. For example, the attached screenshot shows that one cannot make any changes to 
"Number of steps where nobody gains a level". It is always the default value of one even it shows '0'. 
[
Screen Shot 2020-11-11 at 11 13 03 AM
](url)

rbtying:
ah, that's a bug! Do you mind filing that as an issue? Not sure what exactly is happening there.

PR #269, opened 2020-12-12
Support a friend selection policy that disallows point cards

dkgramming:
The group I play with plays with this rule, and I often forget about it when selecting friends.

By the way, thanks for building/maintaining this game!

Screen Shot 2020-12-11 at 9 13 41 PM
Screen Shot 2020-12-11 at 9 13 14 PM
Screen Shot 2020-12-11 at 9 15 07 PM

PR #279, opened 2021-01-18
Update PointCardNotAllowed policy

dkgramming:
I missed a subtle detail when implementing the `PointCardNotAllowed` policy in #269: `king` is supposed to be a valid friend selection when the landlord's rank is `ace`.

I assumed this an uncommonly rule, so I decided to amend the current rule instead of adding a new one. I'd be happy to make a separate one if people actually use the current version.

### DIFF
--- a/frontend/src/Initialize.tsx
+++ b/frontend/src/Initialize.tsx
@@ -51,7 +51,7 @@ const DifficultySettings = (props: IDifficultyProps): JSX.Element => {
               Non-trump cards, except the highest
             </option>
             <option value="PointCardNotAllowed">
-              Non-trump, non-point cards
+              Non-trump, non-point cards, unless the highest
             </option>
           </select>
         </label>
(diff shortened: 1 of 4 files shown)


frontend/src/Initialize.tsx
```diff
@@ -51,7 +51,7 @@ const DifficultySettings = (props: IDifficultyProps): JSX.Element => {
               Non-trump cards, except the highest
             </option>
             <option value="PointCardNotAllowed">
-              Non-trump, non-point cards
+              Non-trump, non-point cards, unless the highest
```
rbtying, line 54: Do you mind making this a bit more clear??
e.g.

`Non-trump, non-point cards (except K when playing A)`

rbtying (commented):
Thanks for submitting this!

It looks like (over the last couple of weeks) this setting is, as you predicted, pretty infrequently selected, so I don't mind if you change the rule rather than adding a new one.

rbtying:
@dkgramming It looks like you need to make some minor lint fixes in `Credits.tsx`; will merge after!

Thanks for working on this

dkgramming:
@rbtying sounds good. Thanks for reviewing it!

rbtying (approved):

PR #280, opened 2021-01-20
Add point progress bar visualization

mwilbz:
## Overview

Hi! I've loved playing Tractor on your website, thanks a lot for writing it! =) One addition I've always thought could be nice is to add a "progress bar" that keeps visual track of each team's points relative to the game's checkpoints. I've implemented it in this PR, but I'm a bit of a frontend noob, so please let me know what changes you'd like to see before approving!

image

## Testing

- Tested locally with both Tractor and Finding Friends modes (screenshot below)
- Ran `yarn prettier --write` and `yarn lint --fix`

### DIFF
--- a/frontend/src/Points.tsx
+++ b/frontend/src/Points.tsx
@@ -24,7 +24,7 @@ const Points = (props: IProps): JSX.Element => {
   const pointsPerPlayer = ObjectUtils.mapValues(props.points, (cards) =>
     ArrayUtils.sum(cards.map((card) => cardLookup[card].points))
   );
-  const { computeScore } = React.useContext(WasmContext);
+  const { computeScore, explainScoring } = React.useContext(WasmContext);
   const totalPointsPlayed = ArrayUtils.sum(Object.values(pointsPerPlayer));
   const nonLandlordPoints = ArrayUtils.sum(
     props.players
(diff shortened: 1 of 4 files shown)


rbtying:
@mwilbz  this is cool! thanks for working on it.

A few things:

1. The number of points needed to pass each level (and make progress in the game) is configurable on a per-game basis. Rather than hardcoding the 20% offsets, you can use an approach similar to the one used in the settings pane: https://github.com/rbtying/shengji/blob/master/frontend/src/ScoringSettings.tsx#L27
2. There is a game option to hide the number of points that the defending team has captured, since that's actually relevant information in the game (and not usually public, at least when I've played in person). We should probably respect that option if it has been set.
3. this is pretty minor, but: if we're using JS to set colors and other styling information, we should be complete about it (i.e. inline the styles as well), rather than having a separate CSS style for the classes

mwilbz:
@rbtying ok thanks for the feedback! I've made the requested changes. The one I wasn't able to test was the `hideLandlordPoints` setting because I couldn't find it in the UI... can you point it out to me?

rbtying:
Hi @mwilbz , thanks for the quick update!

The hide landlord points setting is under difficulty settings
image

frontend/src/Points.tsx
```diff
@@ -93,9 +94,26 @@ const Points = (props: IProps): JSX.Element => {
 
   thresholdStr += ` (next threshold: ${nextThreshold}分)`;
 
+  const { results: scoreTransitions } = explainScoring({
+    params: props.gameScoringParameters,
+    smaller_landlord_team_size: false,
```
rbtying, line 99: ah, probably good to leave a comment here: since you're only looking at point thresholds, the `smaller_landlord_team_size` value doesn't change any of the data that you would care about. But, from the API contract perspective, we actually know whether the landlord has a small team in the parent component.

frontend/src/Points.tsx
```diff
@@ -93,9 +94,26 @@ const Points = (props: IProps): JSX.Element => {
...
+    num_decks: props.numDecks,
+  });
+
   return (
     <div className="points">
       <h2>Points</h2>
+      <ProgressBar
+        checkpoints={scoreTransitions
+          .map((transition) => transition.point_threshold)
+          .filter(
```
mwilbz, line 109: I want to call out this filter -- if there are checkpoint circles really close to the edges it makes the progress bar look bad. And AFAIK this will generally only happen at the 5-point threshold which is somewhat trivial.

mwilbz:
@rbtying thank you! I confirmed that hiding the points works, and I fixed the `smaller_landlord_team_size` argument like you suggested.

rbtying:
This looks great! Thanks a bunch.

If you'd like, you can add a line to the changelog (in `Credits.tsx`) :)

Otherwise, I'll merge this in later today, and it should get deployed tomorrow morning

rbtying (approved):

mwilbz:
Done! Thanks Robert!

PR #288, opened 2021-01-31
Handle races with `arrayMinus` and don't raise an exception

rbtying:


dkgramming:
I think this change may introduced a bug in the autoplay feature. 
autoplay_bug
The correct cards are played, but one them remains as though it were still pending.

PR #305, opened 2021-03-15
Finding friends: add option to allow selecting trumps as a friend

hgbyhgby:
My friends and I play with the rule that any card be selected as a friend, including trumps. Do you mind if I make a change to allow this? 

And thanks so much for creating this! It's awesome!

### DIFF
--- a/core/src/game_state.rs
+++ b/core/src/game_state.rs
@@ -804,7 +804,6 @@ impl ExchangePhase {
             friends.clear();
 
             for friend in friend_set.iter() {
-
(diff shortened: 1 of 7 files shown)


rbtying (commented):
Thanks for adding this! It looks good to me.

Do you mind also adding a line to the changelog, in `Credits.tsx`?

rbtying (approved):

rbtying:
@hgbyhgby thanks! this should go out in the next day or two.

PR #382, opened 2022-09-23
Enhancement: Add a setting for autodraw speed

tliu30:
- added an option to Initialize component (via new AutodrawSpeedSelector component); under Misc Settings
- added new message variant
- added to settings types in both JS & Rust
- changed Draw component to reference autodraw setting, instead of the hardcoded 250ms

### DIFF
--- a/core/src/settings.rs
+++ b/core/src/settings.rs
@@ -346,6 +346,8 @@ pub struct PropagatedState {
     pub(crate) tractor_requirements: TractorRequirements,
     #[serde(default)]
     pub(crate) max_rank: MaxRank,
+    #[serde(default)]
+    pub(crate) autodraw_speed_ms: u32,
 }
 
 impl PropagatedState {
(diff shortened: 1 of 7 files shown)


tliu30:
Love the game, by the way <3 Thanks so much for making this! Have had lots of fun playing this...if this looks good, let me know if there are any other fixes / changes I could help with - would love to help support this great project.

core/src/settings.rs
```diff
@@ -346,6 +346,8 @@ pub struct PropagatedState {
     pub(crate) tractor_requirements: TractorRequirements,
     #[serde(default)]
     pub(crate) max_rank: MaxRank,
+    #[serde(default)]
+    pub(crate) autodraw_speed_ms: u32,
```
rbtying, line 350: minor nit (don't actually make this change):

Ideally this would actually default to `None`, so you can identify the case where the `autodraw_speed_ms` is not set.

However, there's no stateful database, and the number of players drops to zero overnight, so we can probably ignore that (and also drop a bunch of the serde(default) tags)

tliu30:
Hmm... That's a good question.

Since people can individually opt in to autodraw, it makes more sense to make it a player-by-player attribute. Will take a look at that :)

rbtying:
> Hmm... That's a good question.
> 
> Since people can individually opt in to autodraw, it makes more sense to make it a player-by-player attribute. Will take a look at that :)

You can take a look at this commit for hints: https://github.com/rbtying/shengji/pull/365/commits/f968edaf27091553c20fe6afb22a78bab57cf9af

tliu30:
Per comments, abandoning in favor of defining it as a per-user setting (see https://github.com/rbtying/shengji/pull/385)

PR #385, opened 2022-09-24
Enhancement: Add a setting for autodraw speed

tliu30:
- added to player settings 
    - new dropdown in settings pane 
    - added to interface & created a new localstorage setting 
    - propagated to Draw component

### DIFF
--- a/frontend/src/Draw.tsx
+++ b/frontend/src/Draw.tsx
@@ -11,6 +11,7 @@ import InlineCard from "./InlineCard";
 interface IDrawProps {
   state: DrawPhase;
   playDrawCardSound: boolean;
+  autodrawSpeedMs: number | null;
   name: string;
   setTimeout: (fn: () => void, timeout: number) => number;
   clearTimeout: (id: number) => void;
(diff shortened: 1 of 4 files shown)


rbtying:
@tliu30  awesome, thank you!

There’s a couple lint errors — can you run `yarn  prettier —write`, and then `yarn lint`?

Will merge afterwards!!

frontend/src/Draw.tsx
```diff
@@ -92,7 +93,7 @@ class Draw extends React.Component<IDrawProps, IDrawState> {
     ) {
       this.timeout = this.props.setTimeout(() => {
         this.drawCard();
-      }, 250);
+      }, this.props.autodrawSpeedMs || 250);
```
rbtying, line 96: while this is valid JS the typescript compiler doesn’t love it — consider instead

`this.props.autodrawSpeedMs !== null ? this.props.autodrawSpeedMs : 250`

tliu30:
Updated - and thanks for the reminder - should've remembered to check developer notes and run lint / test first - never would've gotten away with being that sloppy at work, haha 😅

Issue #49, opened 2020-04-12
Consider making a listing of "public" rooms that anyone can join


rbtying:
When you create a game, you could mark the room as "public", so that it shows up on a listing prior to joining the room.

jimmyfang94:
Working on this: 

General plan is:
- Add `is_public` to propagated state
- Add action `SetIsPublic` valid on initialized state
- Add message `IsPublicSet`
- Add element under `Difficulty settings`. Something like: "Like Publically?: [Yes/No]"
- Add a table on front page which polls for games with `is_public:true`

@rbtying Thoughts on how best to pull public games from storage? With the current interfaces doesn't seem straightforward but might be missing something.

rbtying:
@jimmyfang94 thanks for thinking about it!

Steps 1-4 makes sense to me, though I'd consider using an enumeration (implementation detail) rather than a boolean
If you look at the implementation of the [state-dump endpoint](https://github.com/rbtying/shengji/blob/master/backend/src/state_dump.rs#L95) you'll see that there's actually already a (hidden) endpoint that enumerates all of the games that you can use for inspiration.

So, we probably need to add:

5. a new endpoint which returns the public rooms as JSON (similar to the current /stats endpoint)
6. some frontend code which polls the endpoint and makes the games available

If we wanted to be extra fancy, we could send the `IsPublicSet=true` message to all clients, which could then immediately poll for the available public games... but I'm not sure this is valuable over a button to refresh the list immediately?

jimmyfang94:
Makes sense! I'll take a stab at it

PR #423, opened 2023-02-25
Implement Game Visibility And Add Public Rooms Panel on Front Page

jimmyfang94:
First pass at: https://github.com/rbtying/shengji/issues/49

- Added a GameVisibility field. Defaults to `unlisted`, but can be toggled to `public`. In the future can be used for private games, etc.
- Added a PublicRoomsPane, with functionality to poll a new API that returns all public games and displays them on the first page. Is polled on first render, and then has a refresh button for ad hoc refreshes afterwards.

Screen Shot 2023-02-24 at 3 34 01 PM
Screen Shot 2023-02-24 at 3 34 12 PM

### DIFF
--- a/backend/src/state_dump.rs
+++ b/backend/src/state_dump.rs
@@ -29,7 +29,7 @@ impl InMemoryStats {
     }
 }
 
-#[derive(Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+#[derive(Serialize, Deserialize)]
 pub struct PublicGameInfo {
     name: String,
     num_players: usize,
@@ -208,6 +208,6 @@ pub async fn public_games(
         }
     }
 
-    public_games.sort();
+    public_games.sort_by_key( |p| (-(p.num_players as isize), p.name.clone()));
     Ok(Json(public_games))
 }
--- a/backend/src/state_dump.rs
+++ b/backend/src/state_dump.rs
@@ -199,15 +199,15 @@ pub async fn public_games(
         if let Ok(versioned_game) = backend_storage.clone().get(room_name.clone()).await {
             if let GameVisibility::Public = versioned_game.game.game_visibility() {
                 if let Ok(name) = String::from_utf8(room_name.clone()) {
-                    public_games.push(PublicGameInfo{
+                    public_games.push(PublicGameInfo {
                         name,
-                        num_players: versioned_game.game.players().len()
+                        num_players: versioned_game.game.players().len(),
                     });
                 }
             }
         }
     }
 
-    public_games.sort_by_key( |p| (-(p.num_players as isize), p.name.clone()));
+    public_games.sort_by_key(|p| (-(p.num_players as isize), p.name.clone()));
     Ok(Json(public_games))
 }
--- a/backend/src/state_dump.rs
+++ b/backend/src/state_dump.rs
@@ -29,6 +29,12 @@ impl InMemoryStats {
     }
 }
 
+#[derive(Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+pub struct PublicGameInfo {
+    name: String,
+    num_players: usize,
+}
+
 pub async fn load_dump_file<S: Storage<VersionedGame, E>, E: Send + std::fmt::Debug>(
     logger: Logger,
     backend_storage: S,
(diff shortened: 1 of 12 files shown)


backend/src/state_dump.rs
```diff
@@ -176,3 +177,27 @@ pub async fn dump_state(
 
     Ok(Json(state_dump))
 }
+
+pub async fn public_games(
+    Extension(backend_storage): Extension<HashMapStorage<VersionedGame>>,
+) -> Result<Json<HashMap<String, GameState>>, &'static str> {
+    let mut public_games: HashMap<String, GameState> = HashMap::new();
```
rbtying, line 184: I think you only need the list of games, right? This might leak information unintentionally.

Perhaps just `Vec<String>` for the names of the rooms? Or, since I see you're using the number of players below, maybe we should just redact all the sensitive values (e.g. as done in https://github.com/rbtying/shengji/blob/d6fee9e36f503d7a47562418f44c8d4e92f484a2/core/src/game_state/mod.rs#L139)

backend/src/state_dump.rs
```diff
@@ -176,3 +177,27 @@ pub async fn dump_state(
...
+
+    backend_storage.clone().prune().await;
+    let keys = backend_storage
+        .clone()
+        .get_all_keys()
+        .await
+        .map_err(|_| "failed to get ongoing games")?;
+    for room_name in keys {
+        if let Ok(versioned_game) = backend_storage.clone().get(room_name.clone()).await {
+            if versioned_game.game.game_visibility() == GameVisibility::Public {
```
rbtying, line 194: I think clippy (`cargo clippy`) will tell you to use

```
if let GameVisibility::Public = versioned_game.game.game_visibility() {

}
```

frontend/src/Initialize.tsx
```diff
@@ -1291,6 +1292,18 @@ const Initialize = (props: IProps): JSX.Element => {
...
+            Game Visibility{" "}
+            <select
+              value={props.state.propagated.game_visibility}
+              onChange={setGameVisibility}
+            >
+              <option value={"Unlisted"}>Unlisted</option>
+              <option value={"Public"}>Public</option>
+            </select>
+          </label>
+        </div>
```
rbtying, line 1306: I think we also need to update the load/save mechanism (e.g. like https://github.com/rbtying/shengji/blob/master/frontend/src/Initialize.tsx#L1045)

rbtying (changes requested):
Very nice work, and thank you again for working on this!

Left some comments inline, but this is pretty close!

backend/src/state_dump.rs
jimmyfang94: Ah makes sense, done. Created a new struct `PublicGameInfo` to hold the two pieces I know. Also moved the sorting logic to backend to make a bit cleaner

backend/src/state_dump.rs
```diff
@@ -28,6 +29,12 @@ impl InMemoryStats {
     }
 }
 
+#[derive(Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+pub struct PublicGameInfo {
+    name: String,
+    num_players: usize,
```
rbtying, line 35: instead of deriving PartialOrd and Ord here (which will sort by default in ascending order of name, then ascending order of number of players), consider using `sort_by_key` in the sort call?

backend/src/state_dump.rs
```diff
@@ -176,3 +183,31 @@ pub async fn dump_state(
...
+                    public_games.push(PublicGameInfo{
+                        name,
+                        num_players: versioned_game.game.players().len()
+                    });
+                }
+            }
+        }
+    }
+
+    public_games.sort();
```
rbtying, line 211: e.g.

```
public_games.sort_by_key(|p| (-(p.num_players as isize), p.name))
```

frontend/src/PublicRoomsPane.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <div className="">
+      <h3>Public Rooms</h3>
+      <div>
+        <p>
+          The games listed below are open to the public. Join them to find new
+          friends to play with!
+        </p>
+        <p>
+          Copy the room name into the input above, fill out your player name,
```
rbtying, line 64: I think we can make these fill it in on click -- we can pass in a `setRoomName` fn as a prop, and then call it in the `onClick` handler for a `PublicRoomRow`. What do you think?

rbtying (commented):
this is looking really good. I'd add a changelog update (in Credits.tsx) so people know about the new feature :)

jimmyfang94:
thanks for the speedy replies! first time playing with rust, fun learning exp

rbtying (approved):

PR #432, opened 2023-06-24
Prevent boolean defaults from overriding saved values

precup:
`show_trick_in_player_order` doesn't get remembered properly on refresh because the default can override it. This tweaks how defaults are handled for loading booleans to respect "off" values.


rbtying:
Oops, this is my bad. Will push this out in the next release (sometime this wknd if I remember!)

rbtying:
THank you for fixing!

PR #442, opened 2023-09-10
Fix bug with double suits being able to outbid double joker

zenador:
Bug: Currently this bidding sequence at rank 2 is possible: 1x2D -> 2xLJ -> 2xHJ -> 2x2S -> 2xHJ -> 2x2H -> 2xHJ etc
Expected behaviour: Double of any suit should not be able to outbid double jokers
Actual behaviour: Double of any suit can outbid double jokers, so bidding can become an infinite loop

This is because in the fallback condition `new_bid.card.suit() > existing_bid.card.suit()`, jokers have a none value or something, so any suit is considered higher, so I added an additional condition before that.

Separately, currently this behaviour is enforced: `The new bid count must have a size of at least 2 in order to be compared by suit ranking`, so you can't have 1x2D -> 1x2S. I've never heard of this condition though, my gaming groups have always allowed that, and this behaviour is not expected from the option description `Joker or higher suit bids to outbid non-joker bids with the same number of cards`. Would it be okay if I add a new option for that and update the text of the current one?

### DIFF
--- a/mechanics/src/bidding.rs
+++ b/mechanics/src/bidding.rs
@@ -526,6 +526,31 @@ mod tests {
                 BidReinforcementPolicy::ReinforceWhileWinning,
                 vec![b!(p, C_2, 3)],
             ),
+            (
+                vec![b!(p, C_2, 1), b!(PlayerID(1), Card::SmallJoker, 2)],
+                BidReinforcementPolicy::ReinforceWhileWinning,
+                vec![
+                    b!(p, C_2, 3),
+                    b!(p, Card::BigJoker, 2),
+                ],
+            ),
+            (
+                vec![b!(p, C_2, 1), b!(PlayerID(1), Card::BigJoker, 2)],
+                BidReinforcementPolicy::ReinforceWhileWinning,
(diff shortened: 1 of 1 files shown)


rbtying:
Hi @zenador , thanks for reporting and fixing! Agreed that this was not intended behavior.

Do you mind adding a test case for this in the `test_valid_bids_joker_or_higher_suit` fn at the end of the file? Happy to merge after.

 With respect to "The new bid count must have a size of at least 2 in order to be compared by suit ranking," -- there's lot of variants of the game, and I believe _most_ of them neither respect suit ordering nor allow single-card-overbidding. But, of course, the goal of this app is to support all play styles, so I'd be happy to accept a pull request that adds an option!

Note that you'll need to update both frontend code (to supply the option) and the core bidding logic (which you're already poking at here).

zenador:
Thanks for the review (and the wonderful project, everything seems very thorough including tests and docs), added new test cases.

Thinking about it again, you're right, I must have misremembered the bidding rule, so I won't follow up with the second part.

PR #452, opened 2024-01-13
Fix rules link

fangherk:
Screenshot 2024-01-12 at 4 54 57 PM

If you hover over the `rules` hyperlink, it redirects to `https://robertying.com/shengji/rules`, which leads to a 404. Map it back to the actual page: `https://robertying.com/shengji/rules.html`.


rbtying:
thank you!

PR #466, opened 2024-12-31
Add Jack Variation

arunsrinivas20:
If the trump number is J and the non-dealer team wins the last trick with a single jack, then the dealer team goes back to rank 2. All rank changes still apply **after** the rank reset occurs. It is an [optional rule](https://en.wikipedia.org/wiki/Sheng_ji#Miscellaneous_optional_rules).

The setting is located under **More Game Settings**.

image

### 4-Player Jack Simulation

The dealer team (A, C) is at J and the non-dealer team (B, D) is at 9. The last trick is won by the non-dealer with a single jack . Without the variation enabled, the dealer team would advance to rank K (they kept the non-dealer team under 40 points). 

image

After clicking **Finish Game**, we will see that the jack variation will kick in. The dealer team's new rank will be 4 (see the chat box in the following screenshot), since they got sent back to rank 2, but they still held the non-dealer team under 40 points. The dealer team remains dealer. 

image

### 5-Player Simulation (1)

In this Finding Friends game, A (the dealer) is at J, B is at 9, C is at A, D is at J, and E is at J. C joins A during the game to form the dealer team. The last trick is won by the non-dealer team with a single jack. Without the variation, the ranks of A and C would be unchanged because they lost the round. B, E, or D would become the dealer. B, E, or D would advance to J, K, and K, respectively. 

image

After clicking **Finish Game**, we will see that the jack variation will kick in. A and C are sent back to rank 2 (see the chat box in the following screenshot), while the other players' ranks match what was described above. 

image

### 5-Player Simulation (2)

In this Finding Friends game, A (the dealer) is at J, B is at J, C is at 2, D is at K, and E is at K. B joins A during the game to form the dealer team. The last trick is won by the non-dealer team with a single jack. Without the variation, the ranks of A and B would advance to Q because they successfully defended. C, D, or E would not become the dealer and would remain at their respective ranks. 

image

After clicking **Finish Game**, we will see that the jack variation will kick in. A and B are set to rank 3 (see the chat box in the following screenshot), while the other players' ranks remain the same. A/B is dealer for the next round. 

image

### DIFF
--- a/core/src/game_state/play_phase.rs
+++ b/core/src/game_state/play_phase.rs
@@ -8,8 +8,8 @@ use shengji_mechanics::deck::Deck;
 use shengji_mechanics::hands::Hands;
 use shengji_mechanics::player::Player;
 use shengji_mechanics::scoring::{compute_level_deltas, next_threshold_reachable, GameScoreResult};
-use shengji_mechanics::trick::{PlayCards, PlayCardsMessage, Trick, TrickEnded, TrickUnit};
-use shengji_mechanics::types::{Card, PlayerID, Rank, Trump};
+use shengji_mechanics::trick::{PlayCards, PlayCardsMessage, Trick, TrickEnded, TrickUnit, PlayedCards};
+use shengji_mechanics::types::{Card, PlayerID, Rank, Trump, Number};
 
 use crate::message::MessageVariant;
 use crate::settings::{
@@ -238,6 +238,8 @@ impl PlayPhase {
             failed_throw_size,
         } = self.trick.complete()?;
 
+        println!("{:?}", self.trick.played_cards());
+
         let kitty_multipler = match self.propagated.kitty_penalty {
             KittyPenalty::Times => 2 * largest_trick_unit_size,
             KittyPenalty::Power => 2usize.pow(largest_trick_unit_size as u32),
@@ -360,6 +362,7 @@ impl PlayPhase {
         landlord: (PlayerID, Rank),
         advancement_policy: AdvancementPolicy,
         max_rank: Rank,
+        lost_on_single_jack: bool,
     ) -> Vec<MessageVariant> {
         let mut msgs = vec![];
 
@@ -373,6 +376,12 @@ impl PlayPhase {
                 };
                 let mut num_advances = 0;
                 let mut was_blocked = false;
+                // If the landlord team lost the final trick with a single jack, then they will go back to Rank 2,
+                // regardless of whether or not they successfully defended or not. If they successfully defended,
+                // then the level bumps will still apply.
+                if is_defending && lost_on_single_jack {
+                    player.reset_rank();
+                };
                 let initial_rank = player.rank();
 
                 for bump_idx in 0..bump {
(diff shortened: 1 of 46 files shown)


arunsrinivas20:
@rbtying can you review this? Let me know if I'm missing something or you need clarification on the setting.

core/src/game_state/play_phase.rs
```diff
@@ -238,6 +238,8 @@ impl PlayPhase {
             failed_throw_size,
         } = self.trick.complete()?;
 
+        println!("{:?}", self.trick.played_cards());
```
rbtying, line 241: do you mind removing this? it'll clutter logs

core/src/game_state/play_phase.rs
```diff
@@ -360,6 +362,7 @@ impl PlayPhase {
         landlord: (PlayerID, Rank),
         advancement_policy: AdvancementPolicy,
         max_rank: Rank,
+        lost_on_single_jack: bool,
```
rbtying, line 365: this is a bit unclear, I think (_which_ team lost on a single jack?).

Perhaps we can just pass in the last trick entirely?

core/src/game_state/play_phase.rs
```diff
@@ -621,6 +631,35 @@ impl PlayPhase {
...
+        }
+
+        let lt_played_cards = last_trick.played_cards();
+        let PlayedCards {
+            cards,
+            ..
+        } = lt_played_cards.iter().find(|pc| pc.id == winner_pid).unwrap();
+
+        // In the jack variation, the last trick must be won with a single (trump) jack
+        return cards.len() == 1 && cards[0].number() == Some(Number::Jack);
```
rbtying, line 660: I've actually heard of multiple variants here, FWIW.

e.g. typically it's fine to win with a pair of J (i.e. triggers J->2), but not to win with HJHJ-J (b/c the winning requires the HJ)

core/src/settings.rs
```diff
@@ -282,6 +282,8 @@ pub struct PropagatedState {
     #[serde(default)]
     pub(crate) hide_throw_halting_player: bool,
     #[serde(default)]
+    pub(crate) jack_variation: bool,
```
rbtying, line 285: Due to the many ways that people play the J->2 variation - do you mind making this an enum so we can add more versions later?

```
BackToTwoSetting {
  #[default]
  Disabled,
  SingleJackToTwo,
  // AceToTwo,
}
```

rbtying (changes requested):
Thank you so much for working on this @arunsrinivas20 ! I'm impressed you figured this all out, I don't think I've seen you contribute before.

A number of nits in the review itself - but the main thing would be to use an enum for the setting rather than a bool because I believe there are additional variants that other folks will want to have also!

arunsrinivas20:
@rbtying thanks for fixing up the linting and format errors! Do you have an example of an `eslint` configuration file for next time (I couldn't run `yarn types ...` locally because of it)? Only `cargo clippy` had issues to fix on my end as a result.

rbtying:
there is a prettier config (so you can do yarn prettier —write), but I think the actual thing is that you want to be using the right version of Node so the yarn install installs the right prettier

(I’m not opposed to someone adding an eslint config but my setup doesn’t seem to need it?)

